PR #19, opened 2019-04-05
fix bug where a mock of a method with body without specified return type was failing

Pmyl:


src/transformer/descriptor/method/methodDeclaration.ts
```diff
@@ -1,10 +1,23 @@
...
 import { GetMethodDescriptor } from "./method";
 import { GetDescriptor } from "../descriptor";
+import {GetReturnTypeFromBody} from "./bodyReturnType";
 
 type Methods = ts.MethodSignature | ts.MethodDeclaration;
 
-export function GetMethodDeclarationDescriptor(node: Methods): ts.Expression {
-	const returnValue: ts.Expression = GetDescriptor(node.type);
-    return GetMethodDescriptor(node.name, returnValue);
+export function GetMethodSignatureDescriptor(node: Methods): ts.Expression {
```
uittorio: I would move this in a different file to be consistent with the idea, 1 SyntaxKind method 1 file. I'm trying to keep it that way.

Pmyl: It makes a lot of sense, my bad.
Done!

PR #22, opened 2019-04-06
make sure methods without return value will not fail the compilation

uittorio:
Nothing really major to mention. to check if there is a return value we need to check the type property in the methodSignature.. Typescript is not really gentle. There is a question mark, so the error was obvious :)


src/transformer/descriptor/method/methodWithReturnType.ts
```diff
@@ -0,0 +1,10 @@
+import * as ts from "typescript";
+import { GetDescriptor } from "../descriptor";
+import { GetMethodDescriptor } from "./method";
+
+type MethodWithReturnType = ts.MethodSignature | ts.MethodDeclaration;
+
+export function GetMethodWithReturnType(node: MethodWithReturnType) {
```
Pmyl: Are you still using this?

PR #34, opened 2019-04-20
feat(createMockList): add createMockList functionality, add typings to framework test

uittorio:
I've added a new feature to createList from mocks.


src/transformer/create-mock-list.ts
```diff
@@ -0,0 +1 @@
+export declare function createMockList<T extends object>(quantity?: number): T[];
```
Pmyl: I would prefer the quantity to be mandatory, is there a reason behind making it optional?

src/transformer/matcher/matcher.ts
```diff
@@ -0,0 +1,34 @@
...
+export function isCreateMockList(declaration: ts.FunctionDeclaration): boolean {
+    return declaration.name && declaration.name.getText() === 'createMockList';
+}
+
+export function isFromTsAutoMock(signature: ts.Signature): boolean {
+    if (!isDeclarationDefined(signature) || !isFunctionDeclaration(signature.declaration)) {
+        return false;
+    }
+
+    if (!isFunctionDeclaration(signature.declaration)) {
```
Pmyl: This is not needed, you have this check in the previous `if`

test/transformer/create-mock-list.test.ts
```diff
@@ -0,0 +1,40 @@
...
+        const properties: Interface[] = createMockList<Interface>();
+        expect(properties.length).toBe(1);
+    });
+
+    it('should create an array of 10 mocks', () => {
+        const properties: Interface[] = createMockList<Interface>(10);
+        expect(properties.length).toBe(10);
+    });
+
+    it('should create an empty array', () => {
```
Pmyl: I would be more explicit like "should create an empty array when the quantity is negative"

src/transformer/create-mock-list.ts
uittorio: I was thinking when you want just to create an array of one... I am not sure is useful. We can remove it for now.

Pmyl: I was thinking that it was creating an empty array, it's not clear what is going to do without input

PR #35, opened 2019-04-26
Refactor(module division) define app structure

uittorio:


src/transformer/matcher/matcher.ts
```diff
@@ -18,8 +18,8 @@ export function isFromTsAutoMock(signature: ts.Signature): boolean {
         return false;
     }
 
-    const createMockTs: string = path.join(__dirname, `src/transformer/create-mock.ts`);
-    const createMockListTs: string = path.join(__dirname, `src/transformer/create-mock-list.ts`);
+    const createMockTs: string = path.join(__dirname, `../create-mock.d.ts`);
```
Pmyl: Why is this targeting the `.d.ts`?

src/transformer/mockDefiner/mockDefiner.ts
```diff
@@ -36,19 +39,31 @@ export class MockDefiner {
...
+                    name: key,
+                    moduleUrl: ModulesImportUrl[key],
+                    identifier: this._createUniqueFileName(key),
+                };
+            });
         }
-        this.currentTsAutoMockImportName = this._neededImportIdentifierPerFile[this._fileName];
+    }
+
+    private _getModuleIdentifier(fileName: string, module: ModuleName): ts.Identifier {
```
Pmyl: Can you move this private method at the bottom?

config/modules/mock/webpack.js
```diff
@@ -0,0 +1,12 @@
+/*globals module, require */
+const path = require("path");
+const merge = require("webpack-merge");
+const base = require("../base/webpack.base");
+
+module.exports = merge(base({
+    module: "mock"
+}), {
+    output: {
+        path: path.resolve(__dirname, "../../../dist")
```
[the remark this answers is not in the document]
uittorio: Because mock goes directly to the dist folder, instead the internal module goes in their folder.

(extensions goes in dist/extensions)
(repository goes in dist/repository)
(mock goes in dist)

src/transformer/matcher/matcher.ts
uittorio: Before I was exporting and publishing also the ts file(create-mock.ts), Now I'm only exporting the d.ts file.

src/transformer/mockDefiner/mockDefiner.ts
uittorio: NOOOOOOOOO!!.. Of course :)

PR #68, opened 2019-07-04
make sure path is the same on windows and unix

uittorio:


uittorio:
Not necessary anymore, because another PR fixed the issue

PR #86, opened 2019-11-02
add global scope and move type reference cache in the scope

uittorio:
This PR includes 2 main things:

1) cache the current NODE so the this reference will be correct
2) refactor the typeReferenceCache to be in the same scope so we control data the be persisted in a single space.


src/transformer/descriptor/identifier/identifier.ts
```diff
@@ -1,8 +1,10 @@
...
+import { IScope } from '../../scope/scope.interface';
 import { GetDescriptor } from '../descriptor';
 import { TypescriptHelper } from '../helper/helper';
 
-export function GetIdentifierDescriptor(node: ts.Identifier): ts.Expression {
+export function GetIdentifierDescriptor(node: ts.Identifier, scope: IScope): ts.Expression {
     const declaration: ts.Declaration = TypescriptHelper.GetDeclarationFromNode(node);
-    return GetDescriptor(declaration);
+    return GetDescriptor(declaration, scope);
+    return GetDescriptor(declaration, scope);
```
Pmyl: It seems you have a repetition here

uittorio: Thanks, However for simplicity, I've included these changes in the generic refactoring PR because the cache will disappear completely.

uittorio:
THis Pr can be closed, because code changes are included in this PR: 
Duplicate of  #87

PR #87, opened 2019-11-03
Generic refactoring

uittorio:
This PR is to improve the functionality for generics. 

Previously generics where not cached so every time you had a generic in your interface we would treat it as another MOCK. This functionality doesn't work well with recursion because the compiler will just die. 

It's still a Work in Progress, the main issue that I have to solve is the extension of interface that is not compatible yet with generic caching. 

Please have a look if you are curios, the code it's still messy because I want to prove that it works for all the scenarios.


uittorio:
Quick update on this PR. The generic functionality is implemented. 
Promises with generics will fail because generics do not work in the same way. Working on a fix

uittorio:
The PR is ready

src/transformer/descriptor/helper/helper.ts
```diff
@@ -70,6 +55,22 @@ export namespace TypescriptHelper {
         }
 
         return TypeChecker().typeToTypeNode(type);
+    }
+
+    export function getTypeParameterOwnerMock(d: ts.Declaration): ts.Declaration {
```
Pmyl: to keep consistency this function should be PascalCase instead of camelCase

src/transformer/descriptor/helper/helper.ts
```diff
@@ -52,16 +19,34 @@ export namespace TypescriptHelper {
...
         const typeChecker: ts.TypeChecker = TypeChecker();
         const symbol: ts.Symbol = typeChecker.getSymbolAtLocation(node.name);
         const declaredType: ts.Type = typeChecker.getDeclaredTypeOfSymbol(symbol);
         return GetDeclarationFromType(declaredType);
     }
 
+    export function GetParameterOfNode(node: ts.EntityName): ts.NodeArray<ts.TypeParameterDeclaration> {
+        const declaration: ts.Declaration = GetDeclarationFromNode(node);
+
+        if (ts.isImportSpecifier(declaration)) {
```
Pmyl: Since now the check `isImportSpecifier` is done in GetDeclarationFromNode is this still required here?

src/transformer/descriptor/helper/helper.ts
```diff
@@ -70,6 +55,22 @@ export namespace TypescriptHelper {
...
+
+    export function getTypeParameterOwnerMock(d: ts.Declaration): ts.Declaration {
+        const typeDeclaration: ts.Declaration = ts.getTypeParameterOwner(d);
 
+        // THIS IS TO FIX A MISSING IMPLEMENTATION IN TYPESCRIPT https://github.com/microsoft/TypeScript/blob/ba5e86f1406f39e89d56d4b32fd6ff8de09a0bf3/src/compiler/utilities.ts#L5138
+        if ((typeDeclaration as Declaration).typeParameters) {
+            return typeDeclaration;
+        }
+
+        const parent: ts.Node = d.parent;
```
Pmyl: this code can be refactored a bit
```
for(let current: ts.Node = d; current = current.parent;) {
  if (current.kind === ts.SyntaxKind.TypeAliasDeclaration) {
    return current as ts.Declaration;
  }
}
```

test/transformer/descriptor/generic/extends.test.ts
```diff
@@ -102,4 +103,145 @@ describe('for generic', () => {
...
+
+        interface C<T> {
+            c: T;
+        }
+
+        interface B<T1, T2, T3> extends A<T1>, C<T2> {
+            b: T3;
+        }
+
+        it('should not consider typescript library', () => {
```
Pmyl: this should not say that, it should say something like "should set the generic value"

test/transformer/create-mock-values.test.ts
```diff
@@ -14,16 +14,4 @@ describe('create-mock-values', () => {
         expect(properties.property).toBe('sss');
         properties.method();
     });
-
-    it('should create an array of mocks with all the value defined', () => {
```
Pmyl: why we're not testing anymore the createMockList with values?

src/transformer/descriptor/helper/helper.ts
uittorio: okay :). Maybe one day we will have consistency across the project!!!

src/transformer/descriptor/helper/helper.ts
uittorio: yes you are right

test/transformer/descriptor/generic/extends.test.ts
uittorio: thanks!

test/transformer/create-mock-values.test.ts
uittorio: I've realised that the test was a duplicate in create-mock-list-values-test

Pmyl: ok!

test/transformer/descriptor/tsLibs/tsLibs.test.ts
```diff
@@ -147,4 +147,59 @@ describe('typescript lib', () => {
...
+        const properties: Interface = createMock<Interface>();
+        const result: string = await properties.method();
+        expect(result).toBe('');
+    });
+
+
+    it('should set a promise resolved for a type mocked directly', async () => {
+        type S<T> = Promise<T>;
+        const properties: S<string> = createMock<S<string>>();
+        createMock<S<number>>();
```
Pmyl: I think this line with just `createMock` is redundant

src/transformer/descriptor/helper/helper.ts
uittorio: I've copied this from the typescript implementation.
https://github.com/microsoft/TypeScript/blob/ba5e86f1406f39e89d56d4b32fd6ff8de09a0bf3/src/compiler/utilities.ts#L5138

 I've tried to remove the code you've wrote but apparently I have to specify also "current;" before the new value each loop.

specifying that value will evaluate every time the expression, so if current is not define it will stop iterate through it and return undefined.

test/transformer/descriptor/tsLibs/tsLibs.test.ts
uittorio: Yes. I wanted to make sure that other mocks do not impact, but we should have separate tests for that

src/transformer/descriptor/helper/helper.ts
Pmyl: it's ok I'm not overly attached to my alternative... but I'm pretty sure it works, it's strange that it doesn't work for you

closing this comment

src/extension/method/provider/provider.ts
```diff
@@ -1,6 +1,6 @@
 import { functionMethod } from './functionMethod';
 // tslint:disable-next-line:no-any
-type Method = (name: string, value: any) => () => any;
+type Method = (name: string, value: () => any) => () => any;
```
Pmyl: This is a breaking change isn't it? Do we want to go through a deprecation phase before or do you think it's ok going forward with this?

Deprecation would be like:
• adding a new type MethodWithDeferredValue
• adding a new method provideMethodWithDeferredValue
• deprecate provideMethod
• change provideMethod wrapping the provided method and calling value() before passing it (to match the old behaviour)

Let me know

uittorio: I think it make sense! Its also a good exercise for us to finally have a library that doesnt do breaking changes immediately and we dont have to move to v2 until we are ready. I have to document the deprecation and explain in which scenario the compiler will die with max call stack

uittorio: done

Issue #89, opened 2019-11-14
Variable Declaration 


uittorio:
In typescript you can define a Variable with the same name as an interface
```ts
var Test: Test;

interface Test {
   a: string;
}

createMock<Test>();
```
This should refere to Test interface and not variable. 

Typescript should always use the interface and not the variable.

**Issue**
The compile fail because its trying to use variable declaration as type;
TsAutoMock doesn't distinguish about multiple declarations and will always use the first one.

Possible solution
A possible solution is to ignore any variable declaration from declarations list.

uittorio:
#83

uittorio:
This issue has been resolved in version 1.3.1

Issue #84, opened 2019-10-30
Implementing VariableDeclaration


joshuavial:
related to #83 - when I setup the following test

```javascript
import { createMock } from 'ts-auto-mock';
import { Connection, Model, Document} from 'mongoose';

describe('for mongoose', () => {
    describe('connection', () => {
        interface Thing extends Document{
          name: string
        }
        interface Interface {
          thing: Model<Thing>
        }

        fit('should set an empty string', () => {
            const mock: Interface = createMock<Interface>();
            console.log(mock.thing);
        });
    });
});

```

It gives the following error
```
Child process failed to process the request: TypeError: Cannot read property '0' of undefined
    at /ts-auto-mock/dist/transformer/index.js:1:6932
```
running it with jest-ts-auto-mock also helpfully reveals 
`NOT IMPLEMENTED VariableDeclaration`

I'm happy to have a crack at implementing VariableDeclaration if that would be useful.

joshuavial:
```javascript
import { createMock } from 'ts-auto-mock';
import { Connection, Model, Document} from 'mongoose';

describe('for mongoose', () => {
    describe('connection', () => {
        interface Thing extends Document{
          name: string
        }
        interface Interface {
          thing: Model<Thing>
        }

        fit('should set an empty string', () => {
            const mock: Interface = createMock<Interface>();
            console.log(mock.thing);
        });
    });
});

```

uittorio:
Hi, I will start working on this issue soon.
The generics implementations to fix the majority of missing implementations of mongoose is in code reivew. 
This issue is a bit tricky because it introduce multiple declaration for the same type
In this instance

```ts 
var Collection: Collection;
interface Collection {}
```
I need to first understand what is this scenario, if its intended by typescript (I believe so) and in that case select the correct declaration :)

uittorio:
This issue has been resolved in version 1.3.1

PR #92, opened 2019-11-16
make sure variable declaration is ignored

uittorio:
This pr is to ignore Variable declarations from declarations list.

There are still a couple of places where we still access to the first declaration without checking if there is not a variable declaration

**TypeParameter.**
```ts
interface Person<TTypeParameter> {

}
```
It will never be a variable declaration
It's possible to have multiple declarations but they have to be the same. I've added a test to prove it in generics/multipleDeclarations.test.ts

**PropertySignature**

Same as above, I've added a test to prove that it still works properties/multipleDeclarations.test.ts


uittorio:
#89 #84

src/transformer/descriptor/helper/helper.ts
```diff
@@ -61,4 +61,10 @@ export namespace TypescriptHelper {
             }
         }
     }
+
+    function GetFirstValidDeclaration(declarations: ts.Declaration[]): ts.Declaration {
+        return declarations.filter((declaration: ts.Declaration) => {
```
Pmyl: you can use `.find` instead of `.filter+[0]`

uittorio: Done

Issue #90, opened 2019-11-14
Recursion in intersection


uittorio:
This Interface should be compiled correctly without giving errors

```ts
interface InterfaceWithIntersection {
    int: InterfaceWithIntersection & B;
 }

 createMock<InterfaceWithIntersection>

```
This interface is valid because an intersection (InterfaceWithIntersection & B) its just the combination of multiple types.
The current compiler is failing because is trying to resolve all the properties of an intersection type causing a maximum call stack error. 

_This is what is happening_
 1) It will first resolve Interface
 2) it will find the property int
 3) it will detect that is an InterfaceWithIntersection
 4) it will resolve all the properties of InterfaceWithIntersection & B.

Because its resolving all the properties of InterfaceWithIntersection it will do keep resolving itself.

The proposed solution is to create a cache declaration for Intersection type, like we do for any other declarations.

The cache declaration will contains 2 declarations and every time the compiler will find an intersectionType it will check the cache first, so it will not resolve all the properties again

With this solution we have to be careful about
 - generics,
 - intersection between primitive types (string)
 - intersection with typescript libraries, like promises

We should create multiple tests for all the scenario above

uittorio:
#83

uittorio:
this issue is resolved by #93

uittorio:
This issue has been resolved in version 1.3.1

PR #93, opened 2019-11-17
Intersection recursive

uittorio:
This PR adds support to intersections that reference to itself.
Example:

```ts
interface Interface {
            a: Interface & {s: string}
}

```

The solution is to cache the intersection types declaration. This solution brings also the benefit to support generics for intersections types.

Note:
Intersection are not fully supported in this library. 
1) the functionality to merge all the sub property doesn't work yet. It's not part of this story, a bug will be created  
2) using the same interface more than once in an intersection will not work correctly. 
If for example the two interfaces will implement 2 different generic values, only the first one will be considered. It's not part of this story, a bug will be created.


src/transformer/mockDefiner/cache/factoryDeclarationListCache.ts
```diff
@@ -0,0 +1,44 @@
+import * as ts from 'typescript';
+
+export interface DeclarationCache {
```
Pmyl: we already have a DeclarationCache class in declarationCache.ts, can we call this or that differently?

uittorio: I've renamed the class to DeclarationListCache to follow the same name conventions and I've renamed the interface to DeclarationListCacheElement.

src/transformer/mockDefiner/cache/factoryDeclarationListCache.ts
```diff
@@ -0,0 +1,44 @@
...
+        return this._find(declarations).key;
+    }
+
+    public has(declarations: ts.Declaration[]): boolean {
+        return !!this._find(declarations);
+    }
+
+    private _find(declarations: ts.Declaration[]): DeclarationCache {
+        return this._cache.find((intersection: DeclarationCache) => {
+            const declarationsCopy: ts.Declaration[] = [...declarations];
```
Pmyl: Can you tell me what this algorithm is doing? It seems like you're checking if `intersection.declarations` contains `declarations`. If I'm correct can you explain the reasoning behind this approach?

uittorio: Now that I'm reading it I'm finding that ArrayA is a subset of ArrayB. I should add the length check at the beginning.
Considering that I will fix with that check I am doing this to make sure that the array have the same elements.
If I don't create the copy and just check of the IndexOf this scenario will result true even if is false

arrayA = ['a', 'b']
arrayB = ['a', 'a']

This is what I'm doing

- Create a copy of arrayA
- find for each Element of ArrayB the element on ArrayA
- remove the element from the ArrayA
- If the arrayA is Empty it means is good

Pmyl: I would write this function somewhere as a utility so you have a place where to write unit tests for it

uittorio: Done. It's actually good having a setup for some unit test.

uittorio: Basically, the last test would not pass using https://jsperf.com/pmyl-compare-array iterative example because all the element in the first array are present in the second array as first index

Pmyl: I've done some more alternatives in https://jsperf.com/pmyl-compare-array/6
I don't really know if it's worth it to change it, I can approve the pr now, let's keep in mind that this could be quicker with some investigation.
I wouldn't do it now since I expect, after reading the new issues, some of this code to change at some point.

uittorio:
this resolve #90

Issue #88, opened 2019-11-14
Type Literal Reference with recursion


uittorio:
this should compile correctly.

```ts 
interface A {
    a: this;

    b: {
        c: string;
    }
}
```

It gives an error on the compiler.

**Reason:**
The current descriptor for typeLiteral is acting as an InterfaceDeclaration.
Because of this it will replace the current this node causing any other this reference to fail.

**Solution:**
We should treat typesLiteral as a different descriptor that will not store the this value.

 #83

uittorio:
#83

uittorio:
This issue has been resolved in version 1.3.1

PR #97, opened 2019-11-20
add test to cover case where interface has a "this" property and a type literal

Pmyl:
This shows that #88 is not relevant, let me know if I'm doing anything wrong here


uittorio:
Yeah, I think I forgot to write the scenario correctly. There is definitely an issue. Did you try instead

```ts 

interface Test {
 a: { b: 2}
 c: this
}
```

Pmyl:
Interesting, the build fails when the "this" property is declared after the type literal... I'll update the test and work on it

Pmyl:
Done, this is now ready for review @uittorio

uittorio:
thanks!!!! I hope it was not too difficult to find the issue!

uittorio:
if you merge remember to use something like this, so it will appear on the changeling
 
fix(this): make sure literal type will not interfere with "this" reference

PR #98, opened 2019-11-21
add support for this in intersections

uittorio:
Hello, After yesterday intersection feature I've found an interesting bug related to "this" in intersections.

We are currently solving the "this" problem storing the Declaration everytime we find a potential Mock.

The problem with this approach is that Intersection is not supported correctly because they are a combination of Declaration.
To follow the same design I would have to store not just A declaration, but the mock declarations that could be a list.

I've realized that we don't need to store a Declaration, we can just store the mock key and everytime we find "this" we can reuse that key.

Let me know what you think :).

This PR also refactor the typeLiteral to use the same method as InterfaceDeclartion and ClassDeclaration that I've introdocued with Intersection feature.

### DIFF
--- a/src/transformer/mockDefiner/mockDefiner.ts
+++ b/src/transformer/mockDefiner/mockDefiner.ts
@@ -84,23 +84,19 @@ export class MockDefiner {
     }
 
     public getMockFactory(declaration: ts.Declaration): ts.Expression {
-        this.setTsAutoMockImportIdentifier();
-
         const key: string = this._getMockFactoryId(declaration);
 
-        return this._getCallGetFactory(key);
+        return this.getMockFactoryByKey(key);
+    }
+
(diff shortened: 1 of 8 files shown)


Pmyl:
> This PR also refactor the typeLiteral to use the same method as InterfaceDeclartion and ClassDeclaration that I've introdocued with Intersection feature.

I hate you but I will review this pr anyway

src/transformer/mockDefiner/mockDefiner.ts
```diff
@@ -91,6 +91,12 @@ export class MockDefiner {
         return this._getCallGetFactory(key);
     }
 
+    public getMockFactoryByKey(key: string): ts.Expression {
```
Pmyl: you could change getMockFactory to call this one

uittorio: Done! thank you!!!!

Issue #104, opened 2019-11-29
Cannot read property for 3rd party types


rahmatrhd:
I am on a Google Apps Script project using typescript and use `@types/google-apps-script` for the types. I want to create a mock object to test an apps script object locally.

```ts
const mockSpreadsheet = createMock<GoogleAppsScript.Spreadsheet.Spreadsheet>();
```

but when running the test (with `mocha`) I got this error message:

```zsh
> npx mocha -r ./tsnode.js **/*.spec.ts

Cannot read property 'kind' of undefined
```

Previously I tried to use an interface I created by myself at it worked. But when I am using interface from apps script (`GoogleAppsScript.Spreadsheet.Spreadsheet`) it returned an error. I am not sure this is the problem with ts-auto-mock or the 3rd party types

uittorio:
@rahmatrhd Hi. I've investigated the issue and I've found out that EnumDeclaration in namespace behave differently.
```ts

declare namespace NameSpace {
    interface WithEnum {
        a: InsideNameSpace;
        b: InsideNameSpaceValue;
        c: OutsideNameSpace;
    }

    enum InsideNameSpaceValue {
        A = '1',
        B = '2',
    }

    enum InsideNameSpace {
        A,
        B
    }
}
```

I'll work on the fix later today or this weekend. Is nothing major :)

Thanks for the feedback

uittorio:
#106 will fix this issue :). I'll let you know when is merged and published. @rahmatrhd  Thanks again  for the feedback.

uittorio:
new version released :). Feel free to close the bug once you are happy with it!

rahmatrhd:
Awesome!! It's working now after I updated to `1.4.1`.
Thanks a lot for the help, @uittorio 🎉

PR #106, opened 2019-11-30
add support for enum in modules and more specific config for playgrou…

uittorio:
I had to move things around the tsconfig to support custom typings for playground tests.
I've also removed some unused custom json tsconfig and move the playground tests in his own folder.


uittorio:
this will fix #104

PR #115, opened 2019-12-03
feat(constructorType): Add constructor type descriptor

Pmyl:
This PR adds constructor type to the handled types.
Since a function works for most (if not all) cases my solution is to use it:

```ts
interface Test {
  recursive: Test;
  recursiveConstructor: new () => Test;
}
const mock = createMock<Test>();
```

will generate something like this

```js
Object.defineProperties(__tsAutoMockObjectReturnValue, {
  recursive: {
    get: function () {
      return _srecursive || (_srecursive = import_Repository.ɵRepository.instance.getFactory("create__test__mock_1")([]));
    }, set: function (_recursive) {
      _srecursive = _recursive;
    }, enumerable: true
  },
  recursiveConstructor: {
    get: function () {
      return _srecursiveConstructor || (_srecursiveConstructor = function () {
        return import_Repository.ɵRepository.instance.getFactory("create__test__mock_1")([]);
      });
    }, set: function (_recursiveConstructor) {
      _srecursiveConstructor = _recursiveConstructor;
    }, enumerable: true
  }
});
```

I've also updated the documentation (moving the constructor type from unsupported to supported), fixed a piece of documentation, and added a karma configuration+npm script to be able to run the playground js without compilation (this will allow creating poc of the solution in js to see if it works before creating the ast).

Let me know if you have any concern

### DIFF
--- a/config/karma/karma.config.transformer.playground.build.js
+++ b/config/karma/karma.config.transformer.playground.build.js
@@ -1,0 +1,7 @@
+const karmaBaseConfig = require('./karma.config.base');
+
+module.exports = function(config) {
+    const karmaConfig = karmaBaseConfig(config, '../../test/playground/**/*.test.js');
+
+    config.set(karmaConfig);
+};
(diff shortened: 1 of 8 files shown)


uittorio:
great!!

PR #116, opened 2019-12-03
feat(constructSignature): add construct signature

Pmyl:
I'm going to treat construct signature in the same way as call signature.

Something doesn't work perfectly but I think it's a non-problem:

```ts
interface Test {
  new (): string;
}
```

is an accepted signature but my implementation can handle only constructions of object-like values.

Let me know if this concerns you

### DIFF
--- a/src/transformer/descriptor/descriptor.ts
+++ b/src/transformer/descriptor/descriptor.ts
@@ -71,8 +71,9 @@ export function GetDescriptor(node: ts.Node, scope: Scope): ts.Expression {
         case ts.SyntaxKind.FunctionType:
             return GetFunctionTypeDescriptor(node as ts.FunctionTypeNode, scope);
         case ts.SyntaxKind.ConstructSignature:
+            return GetFunctionTypeDescriptor(node as ts.ConstructSignatureDeclaration, scope);
         case ts.SyntaxKind.CallSignature:
-            return GetFunctionTypeDescriptor(node as ts.CallSignatureDeclaration | ts.ConstructSignatureDeclaration, scope);
+            return GetFunctionTypeDescriptor(node as ts.CallSignatureDeclaration, scope);
         case ts.SyntaxKind.ArrowFunction:
         case ts.SyntaxKind.FunctionExpression:
             return GetFunctionAssignmentDescriptor(node as ts.ArrowFunction, scope);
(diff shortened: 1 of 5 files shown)


src/transformer/descriptor/properties/properties.ts
```diff
@@ -9,7 +9,10 @@ export function GetProperties(node: ts.Node, scope: Scope): ts.Expression {
     const type: ts.Type = typeChecker.getTypeAtLocation(node);
     const symbols: ts.Symbol[] = typeChecker.getPropertiesOfType(type);
 
-    const signatures: ReadonlyArray<ts.Signature> = typeChecker.getSignaturesOfType(type, SignatureKind.Call);
+    // const signatures: ReadonlyArray<ts.Signature> = typeChecker.getSignaturesOfType(type, SignatureKind.Call);
```
uittorio: Do you mind removing the commented code?

src/transformer/descriptor/descriptor.ts
```diff
@@ -70,8 +70,9 @@ export function GetDescriptor(node: ts.Node, scope: Scope): ts.Expression {
             return GetMethodDeclarationDescriptor(node as ts.MethodDeclaration, scope);
         case ts.SyntaxKind.FunctionType:
             return GetFunctionTypeDescriptor(node as ts.FunctionTypeNode, scope);
+        case ts.SyntaxKind.ConstructSignature:
```
uittorio: I was trying to follow the convention to keep separate each case even if they call the same function

src/transformer/descriptor/descriptor.ts
```diff
@@ -70,8 +70,9 @@ export function GetDescriptor(node: ts.Node, scope: Scope): ts.Expression {
...
         case ts.SyntaxKind.CallSignature:
-            return GetFunctionTypeDescriptor(node as ts.CallSignatureDeclaration, scope);
+            return GetFunctionTypeDescriptor(node as ts.CallSignatureDeclaration | ts.ConstructSignatureDeclaration, scope);
```
uittorio: I think you can remove the OR

uittorio: Ah sorry. I didn't see that you may still want to keep the switch case separated. Once you've done that you can remove the OR

src/transformer/descriptor/properties/properties.ts
Pmyl: Done

src/transformer/descriptor/descriptor.ts
Pmyl: I put the OR there to make sure the typings are correct... if for some reason we change `GetFunctionTypeDescriptor` to only accept `CallSignatureDeclaration` it would give error... if I remove one of the two then there is no check anymore.

Do you want me to split the 2 cases of the switch?

Pmyl: I split the 2 cases

PR #117, opened 2019-12-05
 feat(typeQuery): add support to typeQuery (typeof) to fix #91

### LINKED ISSUE #91: Support for type of
_typeof_ is not currently supported and it would be nice to implement the correct value.

Example
```ts
var obj = {
   a: ''
}
interface Interface {
    typeOf: typeof obj;
 }

```

The result of typeOf should be the var obj  with default values.
A solution would be to generate the literal type on the fly and keep in the correct scope
I am not sure yet about all the implementations and how to solve it because there are scenario that a bit more complex. For example typeof of an enum type.

This issue needs to be investigated.

Pmyl:
This will fix #91 

I've changed how an import of a type `true|false` works, before was behaving in the same way as Boolean, now it works like a normal union picking the first value.

I had to recreate the enum mock even if it wouldn't be needed (the enum is a constant) because there is no way to force the import to stay there. The best solution is for `typeof Enum` to just assign the Enum identifier forcing the import to not disappear. We can do that when https://github.com/microsoft/TypeScript/issues/17552 gets resolved.

To be able to reuse the enum mock I had to create a new parallel process for it, ideally that will be deleted (due to the above statement).

I've changed how we store imports... because I didn't like the loop that we had. Now to get the import identifier of a module we have a map and the imports are agnostic of the fact that are only for internal modules (in this way it's open for custom imports).

I've just covered:
typeof enum
typeof class
typeof function
typeof variable

I can't think of anything else, let me know.

There are two TODOs in this PR, both of them are there because either typescript is missing some features or I don't know how to work with those cases.

### DIFF
--- a/docs/DETAILS.md
+++ b/docs/DETAILS.md
@@ -246,3 +246,28 @@ const mock = new mockType();
 
 mock = // { a: "" }
 ```
+
+## TypeQuery
+```ts
+enum AnEnum {
+    a,
+    b = 'something',
+}
+
+const mock = createMock<typeof AnEnum>();
+mock.a // 0
+mock.b // 'something'
+mock[0] // 'a'
+
+class AClass {
+    a: string
+}
+
+const mockClass = createMock<typeof AClass>();
+new mockClass().a // ''
+
+function AFunction(): number;
+
+const mockFunction = createMock<typeof AFunction>();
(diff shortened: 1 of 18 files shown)


Pmyl:
@uittorio I've added support for typeof variable!

PR #118, opened 2019-12-07
Import support

uittorio:
I've create a couple of test for import alias. @Pmyl you've mentioned that something was not supported. Could you write down the example so I will update the test and fix the issue in this PR? :)


Pmyl:
ImportEqualDeclaration is not supported yet

You will have it for example when using an interface in a namespace without using the entire namespace, so instead of doing:

createMock<Namespace.Interface>();

we can do

import Interface = Namespace.Interface;
createMock<Interface>();



I think we never covered import star and require

uittorio:
Okay, ill create few test :) thanks

test/transformer/descriptor/utils/namespace/namespace.js
```diff
@@ -0,0 +1,10 @@
+"use strict";
+Object.defineProperty(exports, "__esModule", { value: true });
+var NameSpaceInterfaceImport;
+(function (NameSpaceInterfaceImport) {
+    var Enum;
```
Pmyl: Why is this file pushed?

uittorio: Ahah by mistake, now im using git add by commandline and sometimes I forget to exclude the one I dont want! Ill remove it :)

test/transformer/descriptor/create-mock-list-values.test.ts
```diff
@@ -1,41 +0,0 @@
-import { createMockList } from 'ts-auto-mock';
```
Pmyl: Didn’t you remove these tests in a different commit?

test/transformer/descriptor/utils/namespace/namespace.js
uittorio: Removed

test/transformer/descriptor/create-mock-list-values.test.ts
uittorio: I don't think see. the duplicated test is inside test/transformer/create-mock-list-values.test.ts and create-mock-list.test.ts

PR #125, opened 2019-12-28
feat(registerMock): add registerMock functionality to register custom mocks per project

### LINKED ISSUE #12: add ability to define custom mocks for same type (example, Promises)
With the current implementation of the library to mock a promise with your special promise you need to 
``` ts
const mockManuallyEdited = createMock<{ promise: Promise<string>; }>();
mockManuallyEdited.promise = new CustomFakePromise();
(mockManuallyEdited.promise as CustomFakePromise<string>).resolve('data');
``` 
It would be nice to have the ability to define a mock once and it will be used for that type all the time.
``` ts
declare function alwaysUseCustomMock<T extends object, U extends T>(factory: () => U): <T2>(cb: (mock: T2) => T) => Extension<T2, U>;
// usage
alwaysUseCustomMock<Promise<any>, CustomFakePromise<any>>(() => new CustomFakePromise());
```

Pmyl:
This PR adds the possibility to register a custom mock and it contains tests and documentation of every possible usage.
It also adds a new and better method of using extension strategies but I will create a new PR to add documentation for those.

This PR also fixes a usecase where an intersection with a typeof was breaking the build. Now an intersection with a typeof will ignore the typeof when deciding the generated type.

This closes #12

### DIFF
--- a/src/transformer/descriptor/intersection/intersection.ts
+++ b/src/transformer/descriptor/intersection/intersection.ts
@@ -9,12 +9,12 @@ export function GetIntersectionDescriptor(intersectionTypeNode: ts.IntersectionT
     const nodes: ts.Node[] = GetTypes(intersectionTypeNode.types, scope);
 
     const hasInvalidIntersections: boolean = nodes.some((node: ts.Node) => {
-        return TypescriptHelper.IsLiteralOrPrimitive(node);
+        return TypescriptHelper.IsLiteralOrPrimitive(node) || node.kind === ts.SyntaxKind.TypeQuery;
     });
 
     if (hasInvalidIntersections) {
         return GetUndefinedDescriptor();
     }
 
-    return GetMockFactoryCallIntersection(intersectionTypeNode, scope);
+    return GetMockFactoryCallIntersection(intersectionTypeNode, new Scope(scope.depth - 1, scope.currentMockKey));
 }
--- a/src/transformer/descriptor/intersection/intersection.ts
+++ b/src/transformer/descriptor/intersection/intersection.ts
@@ -16,5 +16,5 @@ export function GetIntersectionDescriptor(intersectionTypeNode: ts.IntersectionT
         return GetUndefinedDescriptor();
     }
 
-    return GetMockFactoryCallIntersection(intersectionTypeNode, new Scope(scope.depth - 1, scope.currentMockKey));
+    return GetMockFactoryCallIntersection(intersectionTypeNode, scope);
 }
--- a/src/transformer/descriptor/intersection/intersection.ts
+++ b/src/transformer/descriptor/intersection/intersection.ts
@@ -9,7 +9,7 @@ export function GetIntersectionDescriptor(intersectionTypeNode: ts.IntersectionT
     const nodes: ts.Node[] = GetTypes(intersectionTypeNode.types, scope);
 
     const hasInvalidIntersections: boolean = nodes.some((node: ts.Node) => {
-        return TypescriptHelper.IsLiteralOrPrimitive(node) || node.kind === ts.SyntaxKind.TypeQuery;
+        return TypescriptHelper.IsLiteralOrPrimitive(node) || ts.isTypeQueryNode(node);
     });
 
     if (hasInvalidIntersections) {
--- a/src/transformer/descriptor/typeReference/typeReference.ts
+++ b/src/transformer/descriptor/typeReference/typeReference.ts
@@ -7,14 +7,10 @@ import { GetDescriptor } from '../descriptor';
 import { TypescriptHelper } from '../helper/helper';
 import { GetTypescriptType, IsTypescriptType } from '../tsLibs/typecriptLibs';
 
-function isTypeAlreadyMocked(declaration: ts.Declaration): boolean {
-    return MockDefiner.instance.hasMockForDeclaration(declaration);
-}
-
 export function GetTypeReferenceDescriptor(node: ts.TypeReferenceNode, scope: Scope): ts.Expression {
     const declaration: ts.Declaration = TypescriptHelper.GetDeclarationFromNode(node.typeName);
 
-    if (isTypeAlreadyMocked(declaration)) {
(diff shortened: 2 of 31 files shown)


src/transformer/mock/mock.ts
```diff
@@ -63,3 +67,15 @@ export function getMockForList(nodeToMock: ts.TypeNode, node: ts.CallExpression)
...
 }
+
+export function storeRegisterMock(typeToMock: ts.TypeNode, node: ts.CallExpression): ts.Node {
+    if (!ts.isTypeReferenceNode(typeToMock)) {
+        Logger('RegisterMock').error('registerMock can be used only to mock type references.');
+        return ts.createEmptyStatement();
+    }
+
+    const factory: ts.FunctionExpression = node.arguments[0] as ts.FunctionExpression;
+    MockDefiner.instance.storeRegisterMockFor(TypescriptHelper.GetDeclarationFromNode((typeToMock as ts.TypeReferenceNode).typeName), factory);
```
uittorio: if you want to avoid the casting to ts.TypeReferenceNode you can invert the if statement

You can also returning only once at the end with if else (I honestly don't mind either way).

Up to you :)

src/transformer/descriptor/intersection/intersection.ts
```diff
@@ -9,7 +9,7 @@ export function GetIntersectionDescriptor(intersectionTypeNode: ts.IntersectionT
     const nodes: ts.Node[] = GetTypes(intersectionTypeNode.types, scope);
 
     const hasInvalidIntersections: boolean = nodes.some((node: ts.Node) => {
-        return TypescriptHelper.IsLiteralOrPrimitive(node);
+        return TypescriptHelper.IsLiteralOrPrimitive(node) || node.kind === ts.SyntaxKind.TypeQuery;
```
uittorio: I am trying to start using ts.isType instead of kind check... for example, ts.isTypeQuery(node)

src/transformer/descriptor/typeReference/typeReference.ts
```diff
@@ -9,6 +10,10 @@ import { GetTypescriptType, IsTypescriptType } from '../tsLibs/typecriptLibs';
 export function GetTypeReferenceDescriptor(node: ts.TypeReferenceNode, scope: Scope): ts.Expression {
     const declaration: ts.Declaration = TypescriptHelper.GetDeclarationFromNode(node.typeName);
 
+    if (MockDefiner.instance.hasMockForDeclaration(declaration)) {
```
uittorio: Considering that now we check up front if we already have the mock shall we simplify the checks inside GetMockFactoryCall? I think there are unnecessary check in this scenario. If you look at this line 

https://github.com/Typescript-TDD/ts-auto-mock/blob/323b697d2ded285d39d6d61b4cf46600092bbbf2/src/transformer/mockDefiner/mockDefiner.ts#L181

we will check the declaration in the cache twice.

In the same way the last if statement that use GetMockFactory call could be simplified to not to the check because it will add a new declaration in the cache and create a new mock

src/transformer/mockDefiner/mockDefiner.ts
```diff
@@ -266,6 +292,20 @@ export class MockDefiner {
         return [];
     }
 
+    private _getRegisterMockInFile(sourceFile: ts.SourceFile): ts.Statement[] {
+        if (this._registerMockFactoryRegistrationsPerFile[sourceFile.fileName]) {
+            return this._registerMockFactoryRegistrationsPerFile[sourceFile.fileName]
+                .map((reg: { key: ts.Declaration; factory: ts.Expression }) => {
+                    const key: string = this._registerMockFactoryCache.get(reg.key);
+                    // const key: string = this._factoryCache.get(reg.key);
```
uittorio: I think you just forgot this line. If I understand the intention is good to have the registerFactoryCache separated so we can clean it when need it.

src/transformer/mockDefiner/mockDefiner.ts
```diff
@@ -137,6 +142,23 @@ export class MockDefiner {
         return this._declarationCache.get(declaration);
     }
 
+    public storeRegisterMockFor(declaration: ts.Declaration, factory: ts.FunctionExpression): void {
+        const key: string = this.getDeclarationKeyMap(declaration);
+
+        this._registerMockFactoryCache.set(declaration, key);
+        // this._factoryCache.set(declaration, key);
```
uittorio: I think you just forgot this line. If I understand the intention is good to have the registerFactoryCache separated so we can clean it when need it.

src/transformer/mockDefiner/mockDefiner.ts
Pmyl: removed, it was a leftover

src/transformer/mockDefiner/mockDefiner.ts
Pmyl: removed, it was a leftover

src/transformer/mock/mock.ts
Pmyl: makes sense, I've inverted the if, removed the casting and removed duplication

src/transformer/descriptor/intersection/intersection.ts
Pmyl: interesting, I've change it here, I think in busy places we should still use the check with kind to make it quicker

src/transformer/descriptor/typeReference/typeReference.ts
Pmyl: I've changed few things:
- now in mockFactoryCall.ts there are 2 different ways to get the call, one where you create a new one, one where you use an existing one, this makes sure a check done earlier is not repeated
- in typeReference.ts we make the check and decide which function to use
- in mockDefiner the ifs are removed

The only downside of this is that now there is no check anymore, a call to "GetMockFactoryCall" of an existing mock factory will create a call to the repository of an existing factory. I would say that is ok, everything is under test so it shouldn't be possible to miss it.

PR #126, opened 2019-12-30
add support for default generics on declaration and extensions

uittorio:


uittorio:
Now we are using the type(generics)declarations instead of the typeArguments 

```ts

Interface<A> { // A is the typeDeclaration
 prop: A;
}
createMock<Interface<string>> /// string is the typeArgument

```

with the typeDeclaration its easier to store default generics if not provided

Example
```ts
interface B<T> {
 bProp: T;
}
Interface<A = string> extends B { 
 prop: A;
}

createMock<Interface>();
```

When we check interface B we didn't really store the default value in A so it was not possible to auto default to string.

The code now is checking against typeDeclaration and if the generic has been provided in typeArguments. If not it will store the default generic.

I hope it make sense

PR #127, opened 2019-12-31
feat(intersectionsfunction): add support for function intersections

uittorio:
I've covered another case and removed one casting... yeah!!!!

Let me know if there were any other scenario about intersections and functions!


Pmyl:
Perfect

PR #128, opened 2020-01-02
feat(typeQuery): add support for typeof of an imported module

Pmyl:
This PR adds support to cases like this:

```ts
import * as STAR from './module1';
import REQUIRE = require('./module2');

const module1Mock: typeof STAR = createMock<typeof STAR>();
const module2Mock: typeof REQUIRE = createMock<typeof REQUIRE>();
```

It takes in account **multiple exports**, **export default** (with ad hoc code) and **export equals**.

I've also:
- added some types to make more clear what is happening with properties and signatures removing the confusion that we had before where everything was casted to a different type every few lines
- added a method to get the alias of a symbol safely (if you call the typescript method with a non-alias symbol it will fail)
- fixed a case where creating a mock of a `typeof Enum` was breaking the original Enum
- fixed the mock of `typeof obj.method`
- added some more random casting because typescript typings are weird

**Please Note:** the mock of a module will contain everything that is exported, even interfaces and types that are not available to use.

### DIFF
--- a/src/transformer/descriptor/helper/helper.ts
+++ b/src/transformer/descriptor/helper/helper.ts
@@ -67,9 +67,18 @@ export namespace TypescriptHelper {
         return symbol.escapedName.toString();
     }
 
+    export function GetAliasedSymbolSafe(alias: ts.Symbol): ts.Symbol {
+        return isAlias(alias) ? TypeChecker().getAliasedSymbol(alias) : alias;
+    }
+
     function GetFirstValidDeclaration(declarations: ts.Declaration[]): ts.Declaration {
         return declarations.find((declaration: ts.Declaration) => {
             return !ts.isVariableDeclaration(declaration);
         }) || declarations[0];
+    }
+
+    function isAlias(symbol: ts.Symbol): boolean {
+        // tslint:disable-next-line no-bitwise
+        return !!((symbol.flags & ts.SymbolFlags.Alias) || (symbol.flags & ts.SymbolFlags.AliasExcludes));
     }
 }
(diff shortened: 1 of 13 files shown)


Pmyl:
> 
> 
> Happy with this, just concern about declaration index access. Are you sure there are now ways to have multiple declarations in these scenarios? :)

When do you think an imported symbol could have multiple declarations? I have to be honest I don't remember the case

Pmyl:
@uittorio I've changed the code slightly to fix a problem where a variable with inferred type of the module didn't work.

To be clear, this is the case:
```ts
import * as STAR from './module';

const moduleVariable = STAR;
createMock<typeof moduleVariable>();
```

I've reorganised the code a bit, I think now it makes more sense.

Let me know about the multiple declarations (comment above).

uittorio:
> > Happy with this, just concern about declaration index access. Are you sure there are now ways to have multiple declarations in these scenarios? :)
> 
> When do you think an imported symbol could have multiple declarations? I have to be honest I don't remember the case

multiple declaration exists when you declare a variableDeclaration and an interfaceDeclaration with the same name
```

const Interface: Interface;

interface Interface {
    a: string;
}
```

Or when you declare multiple declaration of the same interface

```
interface Interface {
    a: string;
}
interface Interface {
    b: string
}
```

The declaration[0] in the second scenario is smart enough to get all the properties even if it doesn't look like it contains them. I just had an issue in the past that for some reason there was a variable declaration on top and I couldn't find the right declaration.

Pmyl:
@uittorio those cases work fine for the same reason: I'm getting the declaration and then I'm sending it again to do the process cycle (a.k.a. GetDescriptor) where, if needed, it will gets all the declarations.

PR #130, opened 2020-01-04
Performance tests

uittorio:
**I've just tested that pushing on master directy doesn't work when the branch is protected. I'm looking into another solution**

**I've found another way to push to protected branches.**

An alternative would be to move performance, ui, definitely typed to a different repo so we can push this data without protected branches

This PR will add the ability to run ts auto mock performances locally and automatically every push on master*(see comments at the end)

The code add 2 main functionality:
- performance project
- ui project

The performance project is a simple node js project that given a performance json configuration will run tests against the local version of ts auto mock.

The configuration can be extended and its explained in the readme.

The performance output is a performance.json saved in ts-auo-mock/data/ folder.

I've decided to use data folder because we can save multiple information there (for example definitelytyped json as well)

The UI is in react and depending how has been deployed it will use the local version of the data json or the production version in data/performance.json. 

Please let me know if there is any SHARED functionality that you may use for definitelyTyped so I can move it out from performance

* I did not enable the performance script to each branch yet because it will create a new commit for that branch that eventually will get deleted.

Considering that this functionality is an experiment I would leave it only to the master branch and local development. In the future we can understand if we find this performance data useful or not.

If you really want to enable for your branch you can just change the pipeline on your branch so it will trigger :)


ui/src/app/performance/performanceService/performanceService.tsx
```diff
@@ -0,0 +1,49 @@
...
+export type PerformanceListData = PerformanceBranchName;
+
+export interface IPerformanceService {
+  get(): Promise<PerformanceListData>;
+}
+
+export function PerformanceService(performanceDataUrl: string): IPerformanceService {
+    return {
+        get: (): Promise<PerformanceListData> => {
+            return axios.get(`${performanceDataUrl}performance.json`).then((result: { data: PerformanceListData }) => {
```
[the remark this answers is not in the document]
uittorio: Yes, I don't know why I did it.. I think I was to lazy to parse the response :P

performance/core/exec/execPromise.js
```diff
@@ -0,0 +1,17 @@
+const exec = require('child_process').exec;
+
+function execPromise(command) {
+    return new Promise(function(resolve, reject) {
+        exec(command, (error, stdout) => {
+            if (error) {
+                console.error(error);
```
Pmyl: I know that you hate me... can we remove the console.error?
I have the same identical function in my branch but I don't have the console.error because I don't need it everytime, I actually do something different with the error.

If we have the same function we can eliminate one.

uittorio: Of course I can, I think it was there for no good reason. Ideally we will do something catching the promise no assuming that we log with console :)

performance/core/fs/fileSystem.js
```diff
@@ -0,0 +1,39 @@
...
+const ncp = require('ncp').ncp;
+const del = require('del');
+
+function fileSystem() {
+    return {
+        createFolder: (folder) => {
+            if(!fs.existsSync(folder))
+                fs.mkdirSync(folder);
+        },
+        writeFileAsync(fileName, data) {
```
Pmyl: This method is synchronous (because you use writeFileSync that is synchronous) yet you write "Async" in the name... it's confusing, can we either use the asynchronous version or change the name to Sync?

performance/core/fs/fileSystem.js
```diff
@@ -0,0 +1,39 @@
...
+            return fs.writeFileSync(fileName, data);
+        },
+        readFileAsync(fileName) {
```
Pmyl: Same here, it says Async but it's actually Sync

ui/dist/index.html
```diff
@@ -0,0 +1,11 @@
+<! DOCTYPE html>
```
Pmyl: There are few files in the `ui/dist/` folder, is that on purpose?

uittorio: Yes, this is the UI build and published. I haven't investigated enough but its the easy way to have GitHub Pages to display something. We need the builded file because how GitHub will display the correct website?

Pmyl:
I'm ok with the changes, it's not complete but this is a big step forward! It will allow also an easy integration of definitely typed ui!

PR #136, opened 2020-01-05
feat(DefinitelyTyped): add process with ui to run createMock on every type of DefinitelyTyped repository

Pmyl:


.github/workflows/definitelyTyped.yml
```diff
@@ -0,0 +1,55 @@
...
+on:
+  push:
+    branches:
+      - 'feature/definitelytype'
+    paths:
+      - 'src/**'
+      - 'definitelyTyped/**'
+
+jobs:
+  performance:
```
uittorio: Could you rename the job to definitelyType.. Basically this name is useful in GitHub to understand which pipeline is.

Let's say for example we want to make this pipeline required for a specific branch, with that name we can select it from the list

.github/workflows/definitelyTyped.yml
```diff
@@ -0,0 +1,55 @@
+name: DefinitelyTyped Tests
+
+on:
+  push:
+    branches:
+      - 'feature/definitelytype'
```
uittorio: just as a reminder. I believe you want to change it to the cron job.

.github/workflows/definitelyTyped.yml
```diff
@@ -0,0 +1,55 @@
...
+          git checkout .
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v2.0.0
+        with:
+          commit-message: Add DefinitelyTyped Tests data
+          committer: typescripttdd <typescripttdd@gmail.com>
+          author: typescripttdd <typescripttdd@gmail.com>
+          title: DefinitelyTyped Tests Data on ${{ env.DATE }}
+          branch: definitely-typed-tests-${{ env.DATE }}
+          token: ${{ secrets.GITHUB_TOKEN }}
```
uittorio: Can you change this one to: token: ${{ secrets.GITHUB_TOKEN_TYPESCRIPTTDD }}

If you use the token I just mentioned the PR will trigger a new build on master  otherwise it will not.(And on master is required)

you can see this issue https://github.com/peter-evans/create-pull-request/issues/48 to understand a bit more why.

package.json
```diff
@@ -61,6 +60,7 @@
     "webpack-node-externals": "^1.7.2"
   },
   "dependencies": {
+    "@types/node": "^13.1.1",
```
uittorio: Was this intentional? Because is in the dependencies whoever install ts-auto-mock will also install types/node.

.github/workflows/definitelyTyped.yml
Pmyl: I'm still working on this file to make sure it will ignore the submodule, right now I have to keep the "on push" hook so that my tests runs everytime I push

.github/workflows/definitelyTyped.yml
Pmyl: This pipeline is just to run the definitely typed tests on the master branch, once I put the cron there is no branch to specify I think

.github/workflows/definitelyTyped.yml
Pmyl: done!

.github/workflows/definitelyTyped.yml
uittorio: I did not explain myself correctly.

jobs:
   performance -> definitelyTyped

Pmyl: Ah sorry

PR #169, opened 2020-01-13
perf(emit): decrease emitted code

Pmyl:
This PR decreate emitted code by using minification techniques and by not always using get/set for properties. The emitted code is reduced in average by **35%**.

Use case is mocking the entire `@types/babel-types` module:

**Current version**
626583 characters

**ADD assign properties by always using element access (`m["prop"] = 0`)**
447669 characters

**ADD reduced string identifier for factory (`create__NAME__mock` to `@NAME`)**
434162 characters

**ADD element access only when signature is there, otherwise object literal (`m = { "prop": 0 }`)**
428726 characters

**ADD `undefined -> void 0` and don't assign `void 0` properties**
408600

**ADD `false -> !1` and `true -> !0`**
404922

I've also found that by reducing Repository/Merge/Extension import variable name and methods name to 1-2 characters we could even drop that number to 35k~ but I think that needs more investigation.


src/transformer/mockDefiner/factoryUniqueName.ts
```diff
@@ -4,6 +4,8 @@ import { KeyCounter } from './keyCounter';
 export type PossibleDeclaration = ts.InterfaceDeclaration | ts.ClassDeclaration | ts.TypeAliasDeclaration;
 
 export class FactoryUniqueName {
+    private static AnonymousIdentifierText: string = '*';
```
uittorio: Can we put these values with the other one? Do you think it's a good idea to keep all these "reduced" identifier names in one place?

src/transformer/descriptor/mock/mockProperty.ts
```diff
@@ -2,13 +2,44 @@ import * as ts from 'typescript';
 import { TypescriptCreator } from '../../helper/creator';
```
uittorio: If you can just rename the file mockProperty to mockProperties would be nice

src/transformer/mockDefiner/factoryUniqueName.ts
Pmyl: done!

Pmyl:
@uittorio I've also renamed mockProperty.ts to reflect the new purpose

PR #202, opened 2020-01-29
fix/typescript-types make sure that typescript type name is always returned so it will be possible to convert it to the correct one

uittorio:
I've found a small issue in our typescript conversion.

to understand how we will perform the conversion we were trying to get the name of the type (Promise, Array, ...)

For types like this one we were not able to retrieve the name (InsertPosition)
```ts
 type InsertPosition = "beforebegin" | "afterbegin" | "beforeend" | "afterend";
```
I've added a fix. to resolve the aliasSymbol when symbol is not present.

Looking at the code I realise that any typescript type that we don't support it will be converted to Undefined. Considering that this library now its in a good shape I would like to try to re enable typescript types except the one that we want to convert to their original value.

I think we should do it after we definitely typed issue will reach a 10% errors. Right now we have a 35% of errors.

### DIFF
--- a/src/transformer/descriptor/tsLibs/typescriptLibsTypeAdapter.ts
+++ b/src/transformer/descriptor/tsLibs/typescriptLibsTypeAdapter.ts
@@ -2,17 +2,13 @@ import * as ts from 'typescript';
 import { Scope } from '../../scope/scope';
 import { TypeChecker } from '../../typeChecker/typeChecker';
 import { GetDescriptor } from '../descriptor';
-import { TypescriptHelper } from '../helper/helper';
 import { GetUndefinedDescriptor } from '../undefined/undefined';
 import { TypescriptLibsTypes } from './typescriptLibsTypes';
 
 export function TypescriptLibsTypeAdapter(node: ts.TypeReferenceNode, scope: Scope): ts.Node {
-  const declaration: ts.Declaration = TypescriptHelper.GetDeclarationFromNode(node.typeName);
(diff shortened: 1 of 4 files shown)


src/transformer/descriptor/tsLibs/typescriptLibsTypeAdapter.ts
```diff
@@ -10,7 +10,9 @@ export function TypescriptLibsTypeAdapter(node: ts.TypeReferenceNode, scope: Sco
   const declaration: ts.Declaration = TypescriptHelper.GetDeclarationFromNode(node.typeName);
   const typeChecker: ts.TypeChecker = TypeChecker();
   const type: ts.Type = typeChecker.getTypeAtLocation(declaration);
-  const typeScriptType: TypescriptLibsTypes = TypescriptLibsTypes[type.symbol.name];
+  const name: string = type.symbol?.name || type.aliasSymbol.name;
```
Pmyl: You can use the “safe get aliased symbol” and do .name

It should be somewhere in the helpers

uittorio: thanks, I also forgot to add a test

uittorio: Done, my scenario was a bit different. symbol was undefined.

I've actually found a best way to find the symbol. Have a look.

Issue #210, opened 2020-01-31
For-of-loop in test


wassy92x:
Unfortunately I found another bug.
As soon as I create a mock of an interface, I can't use For-of-loops in my tests.
I always get an exception: "TypeError: Cannot read property 'ɵRepository' of undefined"
If I comment out createMock, the for-of-loop works correctly.

Example:
```
import { createMock } from 'ts-auto-mock';

interface MyInterface {
    a: string;
}

describe('Foo', () => {
    test('some test', () => {
        const myInstance: MyInterface = createMock<MyInterface>();
        const a: string[] = ['hello', 'world'];
        for (const b of a) {
            console.log(b);
        }
    });
});
```

Pmyl:
Could you tell me your configuration to run it? (Webpack, jasmine, ts-loader, etc)

It could be something related to the fact that for..of creates some polyfill emits that for some reason delete our imports.

I’ll look into it unless @uittorio wants to take this

wassy92x:
My test environment is the following:
- jest-cli: 23.6.0
- jest: 23.6.0
- ts-jest: 23.10.5
- ttypescript: 1.5.10
- typescript: 3.7.3

Jest configuration:
```
{
   [...],
    "testMatch": [
      "**/test/**/*.test.js?(x)",
      "**/test/**/*.test.ts?(x)"
    ],
    "moduleDirectories": [
      "node_modules",
      "src/ts",
      "src/js"
    ],
    "moduleFileExtensions": [
      "js",
      "ts",
      "tsx",
      "json"
    ],
    "transform": {
      "^.+\\.tsx?$": "ts-jest",
      "^.+\\.jsx?$": "babel-jest"
    },
    "globals": {
      "ts-jest": {
        "compiler": "ttypescript"
      }
    }
}
```

tsconfig.json
{
  "compilerOptions": {
    "outDir": "./dist/",
    "noImplicitAny": true,
    "module": "es6",
    "target": "es5",
    "jsx": "react",
    "allowJs": true,
    "sourceMap": true,
    "lib": ["es7", "dom"],
    "downlevelIteration": true,
    "allowSyntheticDefaultImports": true,
    "experimentalDecorators": true,
    "moduleResolution": "node",
    "declaration": false,
    "emitDecoratorMetadata": true,
    "skipLibCheck": true,
    "esModuleInterop": true,
    "plugins": [
      {
        "transform": "ts-nameof",
        "type": "raw"
      },
      {
[...]

PR #211, opened 2020-01-31
make sure import module variable are unique

uittorio:
this is a quick fix until we understand completely how we are supposed to create unique variables names in typescript.

I need to create a test that proves unique variables.

NOTE: there are not really unique these variables, if someone use 'ɵRepository' 'ɵMerge' 'ɵExtensions' as variables name will have collissions issues

### DIFF
--- a/src/transformer/mockDefiner/uniqueIdentifier.ts
+++ b/src/transformer/mockDefiner/uniqueIdentifier.ts
@@ -1,5 +1,5 @@
 import * as ts from 'typescript';
 
-export function PrivateIdentifier(text: string) {
-    return ts.createIdentifier(`ɵ${text}`);
+export function PrivateIdentifier(text: string): ts.Identifier {
+  return ts.createIdentifier(`ɵ${text}`);
 }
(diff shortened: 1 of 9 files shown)


src/transformer/mockDefiner/uniqueIdentifier.ts
```diff
@@ -0,0 +1,5 @@
+import * as ts from 'typescript';
+
+export function PrivateIdentifier(text: string): ts.Identifier {
+  return ts.createIdentifier(`ɵ${text}`);
```
Pmyl: don't we have other places where we use that character? could we reuse this function?

test/playground/playground.test.ts
```diff
@@ -10,8 +10,8 @@ import { createMock } from 'ts-auto-mock';
 
 it('should work', () => {
   interface A {
-    a: string;
-  }
+  a: string;
```
Pmyl: let's not touch this file as much as possible

src/transformer/mockDefiner/uniqueIdentifier.ts
uittorio: done :)

test/playground/playground.test.ts
uittorio: done :)

uittorio:
this should fix #210

PR #222, opened 2020-02-05
fix(transformer): export definitely typed for transformer so it will be easier to read the documentation while developing

uittorio:
I've added the export of d ts file for the transformer.

Note: 
I've also moved the default export in a different file, so if we have to swap in the future to a different transformer will be easier. I've tested it locally with jasmine-ts-auto-mock and it works.

Downside of this:

We are exporting plenty of d.ts files because when you set declaration true it will resolve all of the d.ts for the transformer (array, descriptor etc, etc). (Not sure yet if its worth limit it to the public interface and how)

### DIFF
--- a/config/modules/transformer/tsconfig.json
+++ b/config/modules/transformer/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "extends": "../../../tsconfig.base.json",
   "compilerOptions": {
-    "outDir": "../../../dist/transformer"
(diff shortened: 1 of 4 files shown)


uittorio:
I agree, however i dont know how to do it in a smart way!

Pmyl:
@uittorio we'll see together how to do it, in the meantime pull this in master

PR #228, opened 2020-02-08
feat(map): add Map support as a type

Pmyl:
Also changed a linting rule for indentation, remove some code regarding typescript types, fixed a bug when using typeof variable without type and function call initializer and removed descriptor 'case' for function call


config/modules/transformer/webpack.js
```diff
@@ -7,6 +7,7 @@ const base = require("../base/webpack.base");
 module.exports = merge(base({
     tsConfigFile: 'config/modules/transformer/tsconfig.json'
 }), {
+    devtool: 'cheap-module-eval-source-map',
```
uittorio: you should be able to remove this one :)

Pmyl: removed

src/transformer/descriptor/method/bodyReturnType.ts
```diff
@@ -1,23 +1,26 @@
...
 
   if (functionBody.statements) {
     const returnStatement: ts.ReturnStatement = GetReturnStatement(functionBody);
 
     if (returnStatement) {
-      returnValue = GetDescriptor(returnStatement.expression, scope);
+      returnValue = returnStatement.expression;
     } else {
-      returnValue = GetNullDescriptor();
+      returnValue = ts.createNull();
```
uittorio: Should we use GetNullDescriptor or there is a special reason?  I understand that the wrapper GetNullDescriptor is not really useful but we would follow the same standard.

Pmyl: done!

src/transformer/descriptor/method/bodyReturnType.ts
```diff
@@ -1,23 +1,26 @@
...
-export function GetReturnTypeFromBody(node: ts.ArrowFunction | ts.FunctionExpression | ts.MethodDeclaration | ts.FunctionDeclaration, scope: Scope): ts.Expression {
-  let returnValue: ts.Expression;
+export function GetReturnTypeFromBodyDescriptor(node: ts.ArrowFunction | ts.FunctionExpression | ts.MethodDeclaration | ts.FunctionDeclaration, scope: Scope): ts.Expression {
+  return GetDescriptor(GetReturnNodeFromBody(node), scope);
+}
+
+export function GetReturnNodeFromBody(node: ts.ArrowFunction | ts.FunctionExpression | ts.MethodDeclaration | ts.FunctionDeclaration): ts.Node {
+  let returnValue: ts.Node;
 
   const functionBody: ts.FunctionBody = node.body as ts.FunctionBody;
```
uittorio: Because you are touching this file.. could we change this casting adding the type guard function call?? 

In this case I think it will be if (!ts.isArrowFunction(node))

src/transformer/descriptor/type/type.ts
```diff
@@ -62,5 +58,10 @@ export function GetType(node: ts.Node, scope: Scope): ts.Node {
     return GetType(node.type, scope);
   }
 
+  if (ts.isCallExpression(node)) {
```
uittorio: The responsibility of GetType was to resolve a type given a specific node. For example following imports until you find the type. It is useful to understand upfront if a type is a typescript type or to make sure union types are not trying to combine primitive types. I'm not sure 100% resolving the return type of the function would be in the right place here. I don't mind revisiting the scope of GetTypes.

src/transformer/descriptor/method/bodyReturnType.ts
Pmyl: I did something slightly different but I think you will like it

src/transformer/descriptor/type/type.ts
uittorio: I got it wrong. CallExpression is not like a functionType so it make sense to resolve the return type.

src/transformer/descriptor/type/type.ts
```diff
@@ -62,5 +58,10 @@ export function GetType(node: ts.Node, scope: Scope): ts.Node {
...
+    const declaration: ts.Declaration = TypescriptHelper.GetDeclarationFromNode(node.expression);
+    return GetType(GetReturnNodeFromBody(declaration as ts.FunctionDeclaration), scope);
```
uittorio: You may want to add some test for when is not a function declaration .


``` class Test {
a = test()
}

const test = () => {
  return 'ss'
}```

Comment (Pmyl) [reply src/transformer/descriptor/type/type.ts]: no casting anymore, the change is quite big actually

File: src/transformer/descriptor/callExpression/callExpression.ts [hunk 6]
```diff
@@ -0,0 +1,36 @@
...
+  return GetDescriptor(GetCallExpressionType(node), scope);
+}
+
+export function GetCallExpressionType(node: ts.CallExpression): ts.Node {
+  const declaration: ts.Declaration = TypescriptHelper.GetDeclarationFromNode(node.expression);
+
+  return GetFinalFunctionTypeFromDeclaration(node, declaration);
+}
+
+function GetFinalFunctionTypeFromDeclaration(initialNode: ts.Node, node: ts.Node): ts.Node {
```
uittorio: can you use ts.Declaration instead of ts.Node for the second parameter?

PR #233, opened 2020-02-09
feat(set): add Set support as a type

wassy92x:


uittorio:
I am not sure why the gthub action didn't fire. I'll investigate it. I'll merge for now

Issue #238, opened 2020-02-12
TypeError: Cannot read property '0' of undefined


artem1458:
```
enum SOME_ENUM {
    FIRST = 'FIRST',
    SECOND = 'SECOND',
}

interface IBase {
    someField: string;
    anotherField: number;
}

interface IBugInterface extends Record<SOME_ENUM, IBase[]> {}
```
Can not extend Record, when first field in Record is ENUM, if it's string, everything's fine

uittorio:
@artem1458 thanks for reporting this bug.

There is currently an issue extending enum.

I will look into it in the next few days. Is this blocking your development?

artem1458:
> Is this blocking your development

It's not blocking issue, thanks for fast anser!

uittorio:
I've created a couple of PR to prevent the compiler to fail.

One PR will just have a fail safe. This means that it will not mock the Mapped Type (Record) with the correct value but it will just be null. The compiler will not fail.

The second PR will fix the issue but its really bad implementation. Unfortunately I couldn't come up with a better solution because the scenario is really odd. 

I honestly will go for the first solution, fail safe. I am not sure how common is to extend Records with Enums

uittorio:
I've released a new version of ts-auto-mock 1.5.6 where the compiler doesn't fail in this scenario but the type is not supported yet. 

https://typescript-tdd.github.io/ts-auto-mock/types-not-supported

Pmyl:
can we close this one? It doesn't give error anymore and the fact that the type is unsupported is now in the documentation

PR #241, opened 2020-02-16
feat(extendsMappedType): make sure extending mapped types do not prevent to compile

uittorio:
This PR will make sure that if an interface/class extends a mapped type it will not prevent to compile.

#238 

Is a temporary solution until we find a better implementation for this scenario


src/transformer/descriptor/mock/mockProperties.ts
```diff
@@ -8,7 +8,9 @@ import { PropertyLike } from './propertyLike';
 import { SignatureLike } from './signatureLike';
 
 export function GetMockPropertiesFromSymbol(propertiesSymbol: ts.Symbol[], signatures: ReadonlyArray<ts.Signature>, scope: Scope): ts.Expression {
-  const properties: PropertyLike[] = propertiesSymbol.map((prop: ts.Symbol) => prop.declarations[0]) as PropertyLike[];
+  const properties: PropertyLike[] = propertiesSymbol
+    .filter((prop: ts.Symbol) => !!prop.declarations) // Dynamically generated properties (mapped types) do not have declarations
+    .map((prop: ts.Symbol) => prop?.declarations[0]) as PropertyLike[];
```
Pmyl: I think the ? is not needed here

uittorio: Done, thanks :)

PR #263, opened 2020-03-15
build(definitelytyped): ensure d ts without module can be mocked by d…

uittorio:
Definitely typed

this will add the support for definitely typed that are not modules (they don't export).

I've just refactored some code to reuse some functionality. I've created a new transformer specific for definitely typed.

Now is doing a try a catch to all errors. I should really catch only the specific non-modules error (DONE)

Have fun reviewing it. 

Let me know if you have any questions

There are no tests. There is no need. Ts auto mock didn't change. I don't think we want to have test for this transformer.

- [x] You have added tests


definitelyTypedTests/src/transformer/transformer.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    if (isCreateDefinitelyTypedMock(declaration) && ts.isTypeQueryNode(typeQueryNode)) {
+        const typeChecker: ts.TypeChecker = TypeChecker();
+        const typeQuerySymbol: ts.Symbol = typeChecker.getSymbolAtLocation(typeQueryNode.exprName);
+        const typeQuerySymbolDeclaration: ts.ImportEqualsDeclaration = typeQuerySymbol.declarations[0] as ts.ImportEqualsDeclaration;
+        const symbolAlias: ts.Symbol = typeChecker.getSymbolAtLocation(typeQuerySymbolDeclaration.name);
+        const symbol: ts.Symbol = typeChecker.getAliasedSymbol(symbolAlias);
+
+        if (!symbol.declarations) {
+            const moduleName: string =
```
uittorio: Basically this is the scenario when "is not a module".
 It tries to resolve valid declarations from the module and it will create a mock for each.

Pmyl:
It’s good but you’re crazy

PR #301, opened 2020-04-11
enhancement(types): Enable strictNullChecks and throw errors in unhandled situations

martinjlowm:
This PR enables strictNullChecks and catches undefined behavior. For now, all unhandled situations throw errors with a stack trace, enabling users to supply that in an issue.

If desired I can throw in FIXME's in those places as well.


uittorio:
Hi @martinjlowm, I appreciate the effort that you put into this. We definitely want to migrate to strictNullChecks = true. 

I honestly would like a better error Message for each scenario. I also think that few scenario can be fixed in a different way. I'll explain a bit more in each line

src/extension/extensionHandler.ts
```diff
@@ -25,6 +25,10 @@ export class ExtensionHandler<TMock> {
       return extensionOrPropertyName(this._mock);
     }
 
+    if (!maybePropertyHandler) {
```
uittorio, line 28: Considering that function overload it's not helping us enough (In my opinion people should not be able to call this method without that parameter unless they disable type checking. What we could do instead is to create a different method in the ExtensionHandler. @Pmyl Tell me If I am wrong :)

src/options/options.ts
```diff
@@ -7,9 +7,9 @@ export interface TsAutoMockOptions {
   cacheBetweenTests: TsAutoMockCacheOptions;
 }
 
-let options: TsAutoMockOptions = null;
+let options: TsAutoMockOptions | undefined;
```
uittorio, line 10: In reality, the options passed to SetTsAutoMockOptions are a partial of TsAutoMockOptions. To avoid spreading undefined  we could do something like this.

```ts
const options: TsAutoMockOptions = defaultOptions;

export function SetTsAutoMockOptions(_options: Partial<TsAutoMockOptions>): void {
  Object.keys(_options).forEach((key: string ) => {
    options[key] = _options[key];
  });
}
export function GetOptionByKey<T extends keyof TsAutoMockOptions>(optionKey: T): TsAutoMockOptions[T] {
  return options[optionKey];
}
```

src/transformer/base/base.ts
```diff
@@ -42,13 +42,17 @@ function visitNode(node: ts.Node, visitor: Visitor, customFunctions: CustomFunct
...
 
-  if (!isFunctionFromThisLibrary(signature, customFunctions)) {
+  if (!signature || !isFunctionFromThisLibrary(signature, customFunctions)) {
     return node;
   }
 
-  const nodeToMock: ts.TypeNode = node.typeArguments[0];
+  const nodeToMock: ts.TypeNode | undefined = node.typeArguments?.[0];
+
+  if (!nodeToMock) {
```
uittorio, line 53: Again. this would mean calling createMock without a generic. This means disabling the typecheck. If consumers are willing to use it we should throw a more specific error :) Something like. Hei, it looks like you are using {whatever} without a generic? check the documentation at this link

src/transformer/descriptor/helper/helper.ts
```diff
@@ -16,7 +16,11 @@ export namespace TypescriptHelper {
 
   export function GetDeclarationFromNode(node: ts.Node): ts.Declaration {
     const typeChecker: ts.TypeChecker = TypeChecker();
-    const symbol: ts.Symbol = typeChecker.getSymbolAtLocation(node);
+    const symbol: ts.Symbol | undefined = typeChecker.getSymbolAtLocation(node);
```
uittorio, line 19: So, this could happen, true. I am not sure when. We didn't have any issues with this yet. (Cannot read name of undefined). I would either investigate when this could happen or I would put some more useful error message.

src/create-mock.ts
```diff
@@ -1,3 +1,3 @@
 import { DeepPartial } from './partial/deepPartial';
 
-export declare function createMock<T extends object>(values?: DeepPartial<T>): T;
+export declare function createMock<T extends object | undefined>(values?: DeepPartial<T>): T;
```
uittorio, line 3: If you check my comment below I wouldn't set this as undefined unless there is a specific reason

uittorio:
I left a few comments, I didn't check all of them. Ideally, we want to find the cause of the undefined or null first. I'm okay to start putting this error in place but we need a better error message. What do you think?

martinjlowm:
@uittorio thanks for your comments. I will probably find time tonight (CEST time) to iterate on them and hopefully have some adjustments ready by tomorrow.

It's easier with some feedback to determine what the intended outcome is, especially with that in mind that I started exploring the code base two days ago. So, thank you for that! I agree with your proposal regarding better error messages, I just thought an error is better than unexpected behavior and making the type checker happy.

Also, Happy Easter! :)

src/create-mock.ts
martinjlowm: This was solely intended for the type query test covering `createMock<typeof undefined>`, there, `object` isn't enough. The meaning of it, I'm not sure 😛

uittorio:
@martinjlowm HI :) Happy Easter

Oh don't' worry, After 2 days of exploration you have done a great job.

I Agree with you, an Error is better than unexpected behavior but previously if there was an error the stack would be more clear than the new Error, that's why we have to be a bit more specific to be at least clear as before.

src/extension/extensionHandler.ts
martinjlowm: I presume making the property handler (`maybePropertyHandler`) required is the desired behavior in this case?

src/create-mock.ts
uittorio: I would prefer to remove the eslint in the test. The usual behavior is always with an object.

src/options/options.ts
```diff
@@ -7,16 +7,15 @@ export interface TsAutoMockOptions {
   cacheBetweenTests: TsAutoMockCacheOptions;
 }
 
-let options: TsAutoMockOptions = null;
+let tsAutoMockOptions: TsAutoMockOptions = defaultOptions;
 
-export function SetTsAutoMockOptions(_options: TsAutoMockOptions): void {
-  options = _options;
+export function SetTsAutoMockOptions(options?: TsAutoMockOptions): void {
```
uittorio, line 12: You could remove the question mark and do an if statement in the transformer. if(options) { SetTsAUtoMockOptions(options)}

src/extension/extensionHandler.ts
uittorio: it should be required but because is overloading it's not. I am happy with the error. Something similar to the createMock one. "It looks like you are trying to get an extension for {propertyName} without specifying the handler.

src/transformer/descriptor/mock/mockProperties.ts
```diff
@@ -19,21 +19,21 @@ export function GetMockPropertiesFromSymbol(propertiesSymbol: ts.Symbol[], signa
 
 export function GetMockPropertiesFromDeclarations(list: ReadonlyArray<PropertyLike>, signatures: ReadonlyArray<SignatureLike>, scope: Scope): ts.CallExpression {
   const propertiesFilter: PropertyLike[] = list.filter((member: PropertyLike) => {
-    const hasModifiers: boolean = !!member.modifiers;
+    const { modifiers = ts.createNodeArray([]) }: PropertyLike = member;
```
uittorio, line 22: these changes are not necessary. Could we revert these changes in this PR?

src/transformer/mockDefiner/mockDefiner.ts
```diff
@@ -213,8 +232,9 @@ export class MockDefiner {
   private _getMockFactoryIdForIntersections(declarations: ts.Declaration[], intersectionTypeNode: ts.IntersectionTypeNode): string {
     const thisFileName: string = this._fileName;
 
-    if (this._factoryIntersectionCache.has(declarations)) {
-      return this._factoryIntersectionCache.get(declarations);
+    const factoryIntersection: string | undefined = this._factoryIntersectionCache.get(declarations);
```
uittorio, line 235: I preferred the old version. Does it complain typescript? Can we cast it considering that if has works the get should work?

src/transformer/mockDefiner/mockDefiner.ts
```diff
@@ -250,10 +270,12 @@ export class MockDefiner {
     if (this._factoryRegistrationsPerFile[sourceFile.fileName]) {
       return this._factoryRegistrationsPerFile[sourceFile.fileName]
         .map((reg: { key: ts.Declaration; factory: ts.Expression }) => {
-          const key: string = this._factoryCache.get(reg.key);
+          const key: string | undefined = this._factoryCache.get(reg.key);
```
uittorio, line 273: This cannot fail. We should improve the types

src/transformer/mockDefiner/mockDefiner.ts
```diff
@@ -263,10 +285,12 @@ export class MockDefiner {
     if (this._factoryIntersectionsRegistrationsPerFile[sourceFile.fileName]) {
       return this._factoryIntersectionsRegistrationsPerFile[sourceFile.fileName]
         .map((reg: { keys: ts.Declaration[]; factory: ts.Expression }) => {
-          const key: string = this._factoryIntersectionCache.get(reg.keys);
+          const key: string | undefined = this._factoryIntersectionCache.get(reg.keys);
```
uittorio, line 288: Same as above

src/transformer/mockDefiner/mockDefiner.ts
```diff
@@ -276,10 +300,12 @@ export class MockDefiner {
     if (this._registerMockFactoryRegistrationsPerFile[sourceFile.fileName]) {
       return this._registerMockFactoryRegistrationsPerFile[sourceFile.fileName]
         .map((reg: { key: ts.Declaration; factory: ts.Expression }) => {
-          const key: string = this._registerMockFactoryCache.get(reg.key);
+          const key: string | undefined = this._registerMockFactoryCache.get(reg.key);
```
uittorio, line 303: same as above

src/transformer/transformer.ts
```diff
@@ -26,31 +26,33 @@ const transformer:
 export { transformer };
 
 function visitNode(node: ts.CallExpression, declaration: ts.FunctionDeclaration): ts.Node {
-  const nodeToMock: ts.TypeNode = node.typeArguments[0];
+  const nodeToMock: ts.TypeNode | undefined = node.typeArguments?.[0];
```
uittorio, line 29: Considering that this visitNode already passed the check of typeArguments, I would create an interface that extends ts.CallExpression that always as typeArguments. Is that possible?

test/transformer/create-mock-values.test.ts
```diff
@@ -6,9 +6,9 @@ describe('create-mock-values', () => {
   }
 
   interface Interface {
-    property: string;
-    method(): void;
-    subInterface: SubInterface;
+    property: string | null;
```
uittorio, line 9: Why do we need null?

src/transformer/mockDefiner/mockDefiner.ts
martinjlowm: I changed it because `has` doesn't act like a type guard and the `get` would result in `string | undefined` in either case. One thing I missed though, is that the change doesn't consider empty strings - however, I don't think that is a valid value anyway. `has` does check that implicitly.

src/options/options.ts
martinjlowm: That makes sense and I agree, it should be the user's responsibility to pass in a truthy value.

src/transformer/descriptor/mock/mockProperties.ts
martinjlowm: Sure. I guess doing an unnecessary `ts.createNodeArray([])` call was a bit of a stretch. The goal is to type guard modifiers or tell the type checker that we know what we're doing and add an exclamation mark on `modifiers` in the return statement.

src/transformer/transformer.ts
martinjlowm: I guess we can adjust the signature to something like `ts.CallExpression & { typeArguments: NonNullable<ts.CallExpression['typeArguments']> }`

test/transformer/create-mock-values.test.ts
martinjlowm: `property` (and the other properties) of `Interface` is assigned null in the `should create the mock merging the null values provided` test and they are asserted as such too.

src/transformer/transformer.ts
uittorio: I think we could use (node: ts.CallExpression & { typeArguments: NodeArray<TypeNode> })

src/transformer/descriptor/mock/mockProperties.ts
martinjlowm: On that note, the ESLint config disallows the use of `!` so we would have to cast it instead. I think type guarding is cleaner though and more explicit. `isTruthyValue` here can be used in more places than this, however, the requirement of variable typing would be a bit tedious (since `hasModifiers` is now a function).

```
const isTruthyValue = <T>(arg: T): arg is NonNullable<T> => {
  return !!c;
};

const hasModifiers = isTruthyValue;

if (!hasModifiers(member.modifiers)) {
  return true;
}

return member.modifiers.filter((modifier: ts.Modifier) => modifier.kind === ts.SyntaxKind.PrivateKeyword).length === 0;
```

uittorio: I see, let's simply assign modifiers to a variable. 

```ts
const modifiers: ts.NodeArray<ts.Modifier> | undefined = member.modifiers;

    if (IsTypescriptType(member)) { // This is a current workaround to safe fail extends of TypescriptLibs
      return false;
    }

    if (!modifiers) {
      return true;
    }

    return modifiers.filter((modifier: ts.Modifier) => modifier.kind === ts.SyntaxKind.PrivateKeyword).length === 0;
```

src/transformer/mockDefiner/mockDefiner.ts
martinjlowm: Perhaps, we should allow the ESLint `@typecript-eslint/no-non-null-assertion`-rule. That would allow us to do: `this._factoryIntersectionCache.get(declarations)!`.

src/transformer/mockDefiner/mockDefiner.ts
martinjlowm: Disabling `@typecript-eslint/no-non-null-assertion` would allow us to replace this (and the other places) with the following if we can guarantee that the key will be available at all times.

```
const key: string = this._factoryCache.get(reg.key)!;
```

PR #314, opened 2020-05-09
chore(transformer): Enforce definitive signature for getDeclarationKeyMap and add a comment to why we can do that

martinjlowm:
I know it was me that changed this originally, but I've come to realize that it only adds noise to the code.

Now there's a comment that explains why the cast is okay.

~This PR follows #313~


src/transformer/mockDefiner/mockDefiner.ts
```diff
@@ -152,24 +146,22 @@ export class MockDefiner {
...
+
     if (!this._declarationCache.has(declaration)) {
-      const key: string = this._factoryUniqueName.createForDeclaration(declaration as PossibleDeclaration);
+      key = this._factoryUniqueName.createForDeclaration(declaration as PossibleDeclaration);
 
       this._declarationCache.set(declaration, key);
     }
 
-    return this._declarationCache.get(declaration);
+    // NOTE: TypeScript does not support inference through has/get, but we know
```
uittorio, line 158: Instead of creating a let variable, could you just add the casting to the return and keep the code as it was ?

```
return this._declarationCache.get(declaration) as string;

```

PR #322, opened 2020-05-13
fix(union): ensure union type with null or unknown will mock the first type

uittorio:


Pmyl:
This is a breaking change, do we want to have a new major version for this?

PR #325, opened 2020-05-16
chore(transformer): Use the tsconfig.playground.json config for Webpack playground builds

martinjlowm:
Playground builds and tests now both use the same playground TypeScript config.


config/karma/karma.config.base.js
```diff
@@ -2,42 +2,42 @@ const webpackConfig = require('../test/webpack.js');
 const ProcessService = require('../../utils/process/process');
 process.env.CHROME_BIN = require('puppeteer').executablePath();
 
-module.exports = function(config, url) {
-    const processService = ProcessService(process);
-    const debug = processService.getArgument('DEBUG');
-    const disableCache = processService.getArgument('DISABLECACHE');
+module.exports = function(config, url, typescriptConfig) {
```
uittorio, line 5: Hi, I appreciate the effort but it looks more complicated than I've expected. Do you mind if we keep the configuration simpler for now? We could instead pass as an environment variable, as it is currently supported, so build:playground and test:playground will match.


```
//package.json
"test:playground": "karma start config/karma/karma.config.transformer.playground.js DEBUG=true",
"build:playground": "ttsc --project tsconfig.playground.json"
```

uittorio: I don't mind a bit of duplication until we find a better way to organise our tests. I'm open to suggestions

martinjlowm: Fine by me - I don't know if you remember, but my initial problem was that with the overload option, I had to specify it in two places and spend quite a bit of time figuring out that both entries weren't using the same config.

I think some verbosity of what options were enabled/disabled would have helped me then. Obviously, it's a minor thing and being familiar with the project now I'm aware of the behavior. The goal of this was to avoid the same situation for other new contributors.

PR #330, opened 2020-05-22
fix(registerMock): allow to use mocks defined in variables

marcmrf:

- [x] You have added tests

# Purpose

ts-auto-mock places all register mocks calls in top of the file, just after the imports the transfomer generates too. This is really problematic for defining the mock you want to register outside the registerMock callback.

given

```ts
import { registerMock, createMock } from 'ts-auto-mock';

const mock:Interface = { prop: 'value' };

registerMock<Interface>(() => mock);
```

ts-auto-mock will generate the following wrong code

```js
ɵRepository.ɵRepository.instance.registerFactory("@B_1", function () { return mock; });
import { registerMock, createMock } from 'ts-auto-mock';

const mock = { prop: 'value' };
```

The PR tries to address that issue, registering the mock in the same place they're defined instead of the top of the module.

```js
import { registerMock, createMock } from 'ts-auto-mock';

const mock = { prop: 'value' };
ɵRepository.ɵRepository.instance.registerFactory("@B_1", function () { return mock; });

```


Pmyl:
Is _createRegistration method in mockDefiner.ts needed anymore?
(I would have liked to leave the comment in the code but github is acting weird)

uittorio:
@Pmyl _createRegistration its still used for intersection types and standard mocks. So yes, it's still used.

uittorio:
@Pmyl another question for you. With this implementation, we will lose the ability to use register mock in different files in some scenario when `cacheBetweenTests` is enabled

file1
```ts
describe('', () => {
beforeAll(() => {
     registerMock<Interface>(....)
})
})
```


file2
```ts
const mock: Interface = createMock<Interface>()
```

the first file will register the mock and the second file will import it but when the code will run it will fail because the registration has been done in the beforeAll. I remember you've mentioned the registerMock should only be used in a context where all the file has access to it, for example a webpack context or a setupEnvFile for jest.

Do we have any specific documentation for that? I just want to ensure that consumer will be aware when its possible to use registerMock and how.

Pmyl:
@uittorio it’s in the documentation, last paragraph of the RegisterMock page

Pmyl:
This page https://typescript-tdd.github.io/ts-auto-mock/register-mock @uittorio

uittorio:
@Pmyl are we happy to merge this one in?

Pmyl:
@uittorio I’m happy to merge it, are you ok if that usecase won’t work anymore?

uittorio:
@Pmyl Sorry if I got back to his late. 

From what I understand registerMock has some usage limitation.
- registerMock should only be used in common files -> preparation files for your test 
   - with webpack you can use the context file where you fetch all the tests
    - with jest you should use the setup file? We haven't test this one

Assuming that my assumption about the limitation are correct ensuring the ordering of registerMock used with constant initialisation sounds like a good idea.

PR #333, opened 2020-05-23
Feat/random

uittorio:
This PR will add:

- features options configuration
- the ability to enable random values for string, number, and booleans
- tests suite that should enable all features.

The random implementation it's very simple and it doesn't use any library.

@martinjlowm this functionality will impact your PR related to the overload functions. Hopefully, my implementation should simplify the way we tests features. Let me know if you don't understand how its configure


martinjlowm:
I'm a bit concerned with testing features in isolation like this. For simple features like it, it's okay as long as you know that the feature "injection point" is at the leaves of the logic (here it's primitive descriptors that won't lead anywhere). I can imagine (/know 😄) that won't be the general case when more features are added. That was why I proposed the matrix approach in my PR, such that the base functionality is covered with features (even individually) turned on and off.

uittorio:
@martinjlowm I understand your concern. You've raised a really valid point.

There are a few things that we should understand and or agree on for testing strategy.

- In my opinion, adding if statements inside tests make the test suite a bit more difficult to read
- We should be able to run the same type of tests locally compared to the in the CI without changing the configuration manually

I like the idea to use matrix because it would speed up the test, for example, we could run in parallel all the different types of tests. However, I think that this is a different issue

I believe your concern is related to the fact that one feature could break other functionalities. 
How do we solve this? There is a testing approach for that, I don't' remember the name but basically what we could do it's to create a lot of scenarios with feature and no, and instead of asserting against specific properties we just want to make sure that it doesn't break. Duplicating the entire suite it's a bit unnecessary. For example, to have an entire coverage you would have to create a scenario where x number of features are enabled and x number of features are not.

config/utils/features.js
```diff
@@ -0,0 +1,15 @@
+const ProcessService = require('../../utils/process/process');
+
+function DetermineFeaturesFromEnvironment() {
+    const processService = ProcessService(process);
+    const features = processService.getEnvironmentValue('FEATURES')
```
Pmyl, line 5: we don't have linting on this file but I still think we should folow the same rules so... missing semicolon

config/utils/features.js
```diff
@@ -0,0 +1,15 @@
...
+
+    if (features) {
+       return [
+           'random'
+       ]
```
Pmyl, line 10: missing semicolon

config/utils/features.js
```diff
@@ -0,0 +1,15 @@
...
+    }
+
+    return [];
+}
+module.exports = DetermineFeaturesFromEnvironment;
```
Pmyl, line 15: missing new line at the end of the file

src/random/random.ts
```diff
@@ -0,0 +1,13 @@
+export class Random {
+  public static number(): number {
+    return Math.random() * (10000 - -10000) + -10000;
```
Pmyl, line 3: why having `- -` and `+ -`? It makes it weird for me but maybe I'm wrong

src/transformer/descriptor/string/string.ts
```diff
@@ -1,5 +1,13 @@
 import * as ts from 'typescript';
+import { IsTsAutoMockRandomEnabled } from '../../../options/random';
+import { PropertySignatureCache } from '../property/cache';
+import { RandomPropertyAccessor } from '../random/random';
 
 export function GetStringDescriptor(): ts.Expression {
+  if (IsTsAutoMockRandomEnabled()) {
+    const propertyName: ts.PropertyName = PropertySignatureCache.instance.get();
+    const prefix: ts.StringLiteral = ts.createLiteral(`${(propertyName as ts.Identifier).escapedText}`);
```
Pmyl, line 9: I'm probably wrong here but I think I remember we had a method to extract the string from a ts.PropertyName in a safe way (I feel like casting it as ts.Identifier won't work for every case). Is it something we actually have?

config/utils/features.js
uittorio: make sense. We should do a couple of things in another Pull request.

- enable eslint for javascript
- fail the build when there is an error

src/random/random.ts
uittorio: So, you definitely know how Math.random works but I'll write down why I've followed this syntax.

- Math.random -> a number between 0 and 1
- 10000 - -10000 it's always 20000 but it indicates (max and min)
- -10000 it will add the min again to the value

Basically, let's say Math.random returns

- 1,  1 * 20000 => 20000 +- 10000 = 10000
- 0, 0*20000 => 0 +- 10000 = -10000

You can see the range of min and max.

I'll replace the number values with MIN and MAX constant

config/utils/features.js
uittorio: Done :)

uittorio:
so, after a quick conversation with @Pmyl we've decided that we are happy to use the strategy added in this pull request. We'll add a ticket on the board that will address @martinjlowm point regarding running all tests with features on/off. I'll add the ticket and we will discuss it the next monthly update

PR #338, opened 2020-05-25
Add random ENUM values

artem1458:
Added feature random-enums

### DIFF
--- a/src/transformer/descriptor/enum/enumDeclaration.ts
+++ b/src/transformer/descriptor/enum/enumDeclaration.ts
@@ -8,12 +8,12 @@ export function GetEnumDeclarationDescriptor(node: ts.EnumDeclaration): ts.Expre
   const typesList: ts.LiteralType[] = node.members.map((it: ts.EnumMember) => typeChecker.getTypeAtLocation(it)) as ts.LiteralType[];
 
   if (IsTsAutoMockRandomEnabled()) {
-    const nodesList: ts.Expression[] = typesList.map((type: ts.LiteralType) => {
+    const nodesList: ts.Expression[] = typesList.map((type: ts.LiteralType, index: number) => {
       if (type.hasOwnProperty('value')) {
         return ts.createLiteral(type.value);
       }
 
-      return ts.createLiteral(0);
+      return ts.createLiteral(index);
     });
 
     return ts.createCall(RandomPropertyAccessor('enumValue'), [], [...nodesList]);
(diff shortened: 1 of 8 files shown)


Pmyl:
Could you add some unit tests to cover the `const enum` case?
```
const enum Direction {
    Up,
    Down,
    None = 'Nothing'
}

interface Movement {
    direction: Direction;
}
```

Pmyl:
I know I'm pedantic... I just created #339 due to the fact that we currently don't support enums with computed values.
Luckily computed values don't break the unit tests when mocking enums but I'm concerned about this new feature, could you try to add some tests to cover the computed values? I would like for it to not fail, since we don't support them yet they don't have to work completely, as long as the test doesn't break I'm happy with it.
We will think about adding computed values support later.
@uittorio let me know if you agree

src/transformer/descriptor/enum/enumDeclaration.ts
```diff
@@ -1,12 +1,26 @@
...
 
-  if (type.hasOwnProperty('value')) {
-    return ts.createLiteral(type.value);
+  if (IsTsAutoMockRandomEnabled()) {
+    const nodesList: ts.Expression[] = typesList.map((type: ts.LiteralType) => {
+      if (type.hasOwnProperty('value')) {
+        return ts.createLiteral(type.value);
+      }
+
+      return ts.createLiteral(0);
```
uittorio, line 16: I think we can use the index number here :) Instead of 0 we can use the index from the array map

```ts
const nodesList: ts.Expression[] = typesList.map((type: ts.LiteralType, index: number) => {
      if (type.hasOwnProperty('value')) {
        return ts.createLiteral(type.value);
      }

      return ts.createLiteral(index); // What is this scenario??? read below
```

We should add a test for random values for enum that do not have a value. WHen this is happening? It's happening for enum in module declaration. You can find a test for non random in test/transformer/descriptor/enum.test.ts. This is the test that we should have also for random enum

```ts
describe('for enum from a module', () => {
    it('should assign the first enum value', () => {
        const properties: NameSpace.WithEnum = createMock<NameSpace.WithEnum>();

        // We cannot check against the real ENUM because is a d.ts, it cannot be used as a value
        expect(properties.propertyWithoutDefaultValue).toEqual(0);
        expect(properties.propertyWithDefaultValue).toEqual('1');
    });
});
```

uittorio:
Happy with the changes. @Pmyl about computed values I agree, @artem1458 added a test as well, so when we will fix it we'll ensure that the random functionality will work as expected

PR #342, opened 2020-05-25
build(release): configure semantic release and commitizen

### LINKED ISSUE #337: [Feature] Configure semantic release
See https://github.com/Typescript-TDD/ts-auto-mock/issues/335.

@uittorio is ok to use semantic release and I will provide an initial configuration to replace the existing release workflow.

C0ZEN:
@uittorio ok that's it.
I finally added more stuff than I expected.

**Let me explain:**

- Added a bunch of dependencies for semantic release and commitizen
- Added two badges (+20 in style easy 😎)
- Rewrite the release GitHub workflow to use semantic release through the `release` npm script (note that I also added a cache for the dependencies)
- Used a complete custom configuration for the commit convention with commitizen (allow you to choose which type of commit triggers a breaking change and how do you trigger a breaking change)
- Add an npm script `cz` which provide a CLI to write down a commit based on the configuration above (I love it but some prefers to write their commits manually so I did not added a Husky hook to force the commit to run the CLI)
- Add an npm script `commit` to trigger cz like an alias (easy to remind)
- Configure semantic pull request app to follow the same rules of commit - this will add an action to validate or not the merge request based on the commits (you must add the GitHub App to enable it https://github.com/apps/semantic-pull-requests)

**Workflow:**

- Once the master is updated
- The release workflow is triggered
- It runs the `release` npm script
- The script check the commits
- Validate or not
- Update the package version and the changelog
- Publish a tag
- Publish to npm
- Publish a GitHub release
- Add labels to the PR/issues

**Testing:**

It is kinda hard to test. 
The GitHub action release workflow has changed but in case it fails there is no danger since it only run the semantic release.
The semantic release could be tested with the `dryRun` option to `true` but only on the master branch.
If you want to enable this option to make sure that everything is ok then I shall edit this PR before merging it into master.

**What to do next:**

- Update the type of commits to your own liking
- Update the type of commits triggering a breaking change to your own liking
- Configure the prerelease/beta/next versions if you need to
- Could add a Husky hook to check if the commit follow the commitizen rules
- Could add a Husky hook to use the cz CLI instead of pure text messages

Fixes #337.


package.json
```diff
@@ -40,16 +41,23 @@
...
     "@types/jasmine": "^3.5.10",
     "@types/node": "^13.13.5",
     "@typescript-eslint/eslint-plugin": "^2.31.0",
     "@typescript-eslint/eslint-plugin-tslint": "^2.31.0",
     "@typescript-eslint/parser": "^2.31.0",
     "awesome-typescript-loader": "^5.2.1",
     "clean-webpack-plugin": "^3.0.0",
+    "commitizen": "^4.1.2",
+    "conventional-changelog-angular": "^5.0.10",
     "conventional-changelog-cli": "^2.0.34",
```
uittorio, line 56: Do we still need conventional-changelog-angular and conventional-changelog-cli?

.releaserc.json
```diff
@@ -0,0 +1,74 @@
...
+            "BREAKING CHANGE",
+            "BREAKING CHANGES"
+          ]
+        },
+        "writerOpts": {
+          "commitsSort": [
+            "subject",
+            "scope"
+          ],
+          "owner": "uittorio"
```
uittorio, line 33: Can we use another username? I would like to use typescripttdd.

package.json
```diff
@@ -21,9 +21,10 @@
     "test:features": "cross-env CACHE=true FEATURES=true karma start config/karma/karma.config.features.js",
     "test:playground": "karma start config/karma/karma.config.transformer.playground.js",
     "test:playground:build": "karma start config/karma/karma.config.transformer.playground.build.js",
-    "version": "conventional-changelog -p angular -i CHANGELOG.md -s -r -0",
-    "copyForPublish": "cp -rf package.json README.md CHANGELOG.md dist",
-    "preparePublish": "npm run build && npm run copyForPublish"
+    "release": "semantic-release",
```
uittorio, line 24: I read the documentation briefly. This command -> semantic-release -> will automatically bump the version, create the tags and canghe the changelog file right and then it will create a commit `ci(release): new version ${nextRelease.version}`?

When should be using commitmsg and cz?

.releaserc.json
C0ZEN: @typescripttdd is a GitHub account?
If so, I will make the change.

uittorio: Yes it's a github account I've created to prevent all my stats to be changed on github. This account push every night in this repo :P

package.json
C0ZEN: I will try to delete those just in case, you gave me a doubt.

package.json
C0ZEN: The `release` script behavior is based on the `.releaserc.json` configuration file.
So yes indeed, it does all the things you said plus it create an npm release and a GitHub release.

As indicated in the "What to do next", there is one possible feature to add to check that the commit messages are well written. This is important since commitizen is configured to define how do you wish to write the commits and semantic uses it to define then based on the commits if the next release should be a patch, a minor or a major version. The `commitmsg` in this case is used but as now it can be removed since I did not configure the hook.

Finally, the `cz` script is a wonderful CI tool once again based on the commitizen configuration provided to write the commits without possible errors. This is a mix between manual commit messages and linted commit messages with a hook. People used to the commitizen convention are well aware of the cz script to use as an alternative and is very welcoming to new contributors since they just do not need to read the contributing guide to know how to right a good commit since the CLI got them covered.

C0ZEN: Removed all the configuration related to the git hook for now.

.releaserc.json
C0ZEN: Done.

package.json
C0ZEN: Removed these two dependencies.
The angular package was used as a preset and this was a mistake because the configuration should be a custom one with commitizen.
Thank you for pointing that out and you should be aware that I change the Angular preset (default) by the config option using commitizen.
This should work but once again, the dry run won't let me go further enough to test that so a dry run on the master branch will tell us if it's ok for sure.

C0ZEN:
@uittorio all good.

.github/workflows/release.yml
```diff
@@ -1,52 +1,41 @@
 name: Release
-
 on:
-  release:
-    types: [published]
-
+  push:
```
Pmyl, line 3: I'm not overly happy about creating a new version every push. There are pushes where you don't need versioning (most of `docs`, `chore`, `style` kind of commit) and there are cases where it's better to include multiple breaking changes in only one major version. This library is fairly new, mainly worked by two people and we are already at version 2. Now that other people wants to contribute (big thank you btw) I'm concerned that by adding a new major version every breaking change it will spiral out of control.
Besides "convenience" is there any other reason for doing this?

C0ZEN: Yes there are alternatives @Pmyl.
From my experience, having a continuous deployment is very good.

**Solution 1 (bad design and not possible but I share it anyway):**
You could have a dedicated branch for example which is the stable version.
Nevertheless semantic-release try to create a standard on how we should handle versioning.
For example, the `branches` option allow only very specific branches.
So this branch would obviously be: master.

**Solution 2 (good by design - my favorite):**
Follow the Gitflow and add a develop branch which will become the entry point of your merge request (making it even the default branch).
That way, you can do what you want with code, make tons of breaking change and once you are ok to deliver it you merge it in master.

**Solution 3 (meh):**
Change the release workflow to no longer be automated.
Note that I recommend you to still create the release on a CI.
Then you should find a way to trigger the release by hand (never done that but I am pretty sure that you just can with a hack).
See this [feature request to allow manual trigger](https://github.community/t5/GitHub-Actions/GitHub-Actions-Manual-Trigger-Approvals/m-p/31517#M813).
See this [hack to do it](https://dev.to/s_abderemane/manual-trigger-with-github-actions-279e).

**Solution 4 (meh):**
Change the release workflow to no longer be automated.
Run the release by hand locally (should change the semantic option `ci` to false then).

C0ZEN: @Pmyl `docs`, `chore` and `style` types are configured to not trigger a release.
As I said, the configuration I provided allow you to define which types of commit you want to allow and which type of release it should trigger.

See the configuration:

```
"releaseRules":[
   {
      "type":"feat",
      "release":"minor"
   },
   {
      "type":"fix",
      "release":"patch"
   },
   {
      "type":"perf",
      "release":"patch"
   }
]
```

Pmyl: I like solution 2!
But I'll investigate a bit more, I'll let you know asap, sorry

uittorio: I still think that there is nothing wrong about releasing every push, If we are pushing too many breaking changes and indeed creating too many versions that means that the library its not stable enough and the issue is somewhere else. As @C0ZEN pointed out we can configure the release rules to not trigger for chore, docs etc etc.

I'm also okay to move forward solution 2 with a manual process to merge into master when you are ready. However this is has a similar impact of what we currently have right now (Solution 3).

uittorio: trying to have an automated CD its not only useful to avoid manual process but its a mentality approach that will ensure that we produce quality code without hesitating. We should trust what we write and the infrastructure and tests that we run all the time. Numbers are free, if we reach version 20.0.1 its okay. Automated CD will improve our CI because we want to avoid mistakes.

uittorio: @Pmyl I understand your point and I'm relaxed with whatever solution we apply. Both of them have pro and cons in my view :)

Pmyl: Automated CD works well on an application, a library is fundamentally different. CD on an application means seamless transition for the user, CD on a library means user having to update it potentially every hour instead of at fixed rate. I don't agree with `If we are pushing too many breaking changes [...] that means that the library its not stable enough`, there are many different reasons why breaking changes are needed, it doesn't always mean that the code is not stable. Change in requirements, minimum typescript version, also adding new supported type is in fact a breaking change, it may be that some users were relying on that being always undefined.

I like reading all the changes in a minor/major version of a package, it seems people is working on it. Having a version per bug/feature clutters the changelog and makes it less appealing to me.
e.g. If we approve 4 PRs in half an hour we get 4 new versions, I think it's worse than having one with the 4 commits written in it.

Going back to CD, even with CD you tend to work in sprints/timeboxes where at the end you have a "release note" with everything released. That is what I would like to have here as much as possible.

Let me think about it anyway

C0ZEN: Just so you know but I agree with @Pmyl.
A minimum of control of the releases is from my opinion required and you should be able to decide or not to release a new version.
That is why the Gitflow is pretty common nowadays and is a good fit for both your needs.
Only merging a branch to create a release: this is not seamless but this is not grunt work either.

uittorio: @Pmyl @C0ZEN Happy to go with solution 2, however I keep my mind on automated release on each merge to master for filtered commits messages. These are my reason :)

`CD on a library means user having to update it potentially every hour instead of at fixed rate.`
I see your point, it will create more noise? Yes. User don't have to update every time there is a change. they can just check the releases and decide to update or not.

` there are many different reasons why breaking changes are needed, it doesn't always mean that the code is not stable`

In this library we are following a policy of backwards compatibility for features, we haven't decided yet how long it should be. The point is that there shouldn't be too many.
I agree with you, for typescript updates we don't have a backwards compatibility functionality.


`I like reading all the changes in a minor/major version of a package, it seems people is working on it. Having a version per bug/feature clutters the changelog and makes it less appealing to me.`

I agree with you about readability. On the other hand you could say that it's easier to understand that one version change one thing instead of one version changed 50 things. 
About `it seems people is working on it`. If you release every day it looks like people are working on it anyway, event better, faster, more frequently. 


`if we approve 4 PRs in half an hour we get 4 new versions, I think it's worse than having one with the 4 commits written in it.`
In my opinion its better because as a consumer you can decide to only have one feature and not the entire block. A transition for a consumer its match smoother if they can update one bit at the time instead of plenty of things.

uittorio:
After our conversation what are our next steps?
- create a develop branch
- make develop default and ensure that behaves like now (test pipeline, protected)
-  Review this MR and ensure that the release will work

Anything I'm missing?
@Pmyl @C0ZEN

C0ZEN:
@uittorio eventually update the documentation for the release process and the Gitflow convention.
Also, we need a commit to trigger the release workflow (based on my commits, none will trigger a new version).
Remember that the dryRun option is still enabled so once we are sure that the process is good we should create a new PR to change this option to false.

Did you add the GitHub app for the release convention check?

C0ZEN:
At least this will not look like this shit: https://github.com/eslint/eslint/blob/master/CHANGELOG.md

uittorio:
@C0ZEN Which plugin is -> release convention check?

C0ZEN:
@uittorio https://github.com/apps/semantic-pull-requests

uittorio:
> @uittorio https://github.com/apps/semantic-pull-requests

Oh wow thats beautiful! when I've added conventional-changelog I wondered how could be possible to prevent bad commits in PR.. this is the answer!! haha.. 

Let's wait @Pmyl feedback and I'll activate the plugin once we are ready!!

Great work @C0ZEN , and thanks for sharing your knowledge regarding good practises. Really appreciate.

C0ZEN:
@uittorio you are welcome. Plus by asking some questions you made me check and read more documentation pointing out that I could do better so it's a win-win.

C0ZEN:
I may have forgotten to add some files in the npm publish step.
Let me check that.

C0ZEN:
I tested the configuration without the angular preset the release notes are just pure shit. I will look into it but I do not wish to merge this until I found a good workaround.
The preset was obviously leveraging the release notes by grouping them by commit type.
Thus current configuration only list ALL commits which is completely useless.

Pmyl:
I didn't change my mind, I would like to go with the Gitflow solution if there is a way.
If you need anything from me let me know

uittorio:
@Pmyl cool. 

I'll create a PR that changes the existing workflow to trigger against master.

uittorio:
@C0ZEN I've created the PR to change the workflow to use develop as branch.

uittorio:
@C0ZEN whenever you are ready these are the steps I've done.

- [x]  create develop branch
- [x]  change workflow to use develop
- [x]  set develop as default
- [x]  add plugin convention check

We still need to:
- [] update Pull request removing dryRun
- [] update pull request to merge to develop instead
- [] add some documentation somewhere regarding release process
- [] create one pull request against develop to test all the workflow
- [] create one pull request against develop to test that the commit will go to the changelog once its merged to mast er
- [] create one pull request against develop to test that a commit will not go the changelog once its merged to master


I'll wait for your changes :)

C0ZEN:
We still need to:

- [ ] ~~update this pull request removing dryRun~~
- [ ] update this pull request to merge to develop instead (@uittorio I can not do it due to rights limitation)
- [ ] add some documentation somewhere regarding release process (@uittorio better if you do it since you know the project from A to Z)
- [ ] ~~create one pull request against develop to test all the workflow~~
- [ ] ~~create one pull request against develop to test that the commit will go to the changelog once its merged to master~~
- [ ] ~~create one pull request against develop to test that a commit will not go the changelog once its merged to master~~


**Note:**
I want to use the dry run option when merging to make sure that everything is ok (the release workflow should be triggered once the develop is merged in the master and with a good commit we should see in the logs if everything is set as expected - this is way safer to avoid risking a bad release).
This test could not be done in the develop due to semantic release limitation.

uittorio:
@C0ZEN We are good to go.. I'll merge this one to develop and then I'll create a PR to merge develop to master and see what happens.

PR #355, opened 2020-05-31
merge develop into master

uittorio:


C0ZEN:
@uittorio that is precisely the moment you start to like the cz 😆 
It avoid bad commits!

C0ZEN:
Anyway, since your commit is chore you can merge the PR with the admin rights this is not a problem.

PR #379, opened 2020-06-13
chore(build): Switch to commonjs2 libraryTarget

martinjlowm:
Emit `__esModule` indicators for ES modules to distinguish between
them and CommonJS modules.

If libraryTarget is set to `commonjs`, the bundle does not match the interface
provided by the type declarations because `default` imports end up as
`module.exports`, e.g.:

```
import transformer from 'ts-auto-mock/transformer'; // commonjs -> { default: [Function (anonymous)] }
import transformer from 'ts-auto-mock/transformer'; // commonjs2 -> [Function (anonymous)] ✓
```


uittorio:
I'll have a look!! i didnt know about the existence of commonjs2.

Ignore the failing automerge.  I need to do some cleanup, its not necessary for this pr

uittorio:
I'm wondering when commonjs should be use instead of commonjs2 now that I understand the issue.

martinjlowm:
The documentation is here: https://webpack.js.org/configuration/output/#module-definition-systems 

The explanation is a bit vague, but afaik the difference is really that `commonjs` supports CommonJS modules only, whereas `commonjs2` adds a compatibility layer between ES (and maybe other) modules and CommonJS modules. Since this project is built using TypeScript (and ES modules) we need that compatibility.

I stumbled over it after I upgraded ts-jest, which for some reason throws away custom transformers defined in tsconfig.json, thereby no longer supporting the ttypescript interface. Not sure if it will be fixed, but I went ahead and migrated to the ts-jest `astTransformer`-option which required me to specify transformers with a new interface, i.e.:

```
import transformer from 'ts-auto-mock/transformer';
import { ConfigSet } from 'ts-jest/dist/config/config-set';

export const factory = (cs: ConfigSet) => {
  return transformer(cs.tsCompiler.program, {
    cacheBetweenTests: false,
    debug: true,
    features: ['overloads'],
  });
};
```

Here, the types would import `transformer` correctly, however at runtime, it would become `{ default: transformer }(...)` because we weren't emitting the `__esModule` indicator which TypeScript assumes the user will provide.

PR #388, opened 2020-06-20
chore(build): Split dist packaging into its own script

martinjlowm:
This adds convenience in local development in other projects like so:

```
git clone https://github.com/Typescript-TDD/ts-auto-mock.git
git checkout <custom-branch>
npm ci

npm run build
npm run dist:collect
```

or through symlinks to a local copy (ts-auto-mock/dist) with custom changes.

---

Personally, I use Yarn 2 and the `exec`-protocol to execute the script above allowing me to depend on a patched version of ts-auto-mock.


package.json
```diff
@@ -21,7 +21,8 @@
     "test:features": "cross-env CACHE=true FEATURES=true karma start config/karma/karma.config.features.js",
     "test:playground": "karma start config/karma/karma.config.transformer.playground.js",
     "test:playground:build": "karma start config/karma/karma.config.transformer.playground.build.js",
-    "release": "cp -r package.json package-lock.json README.md dist && semantic-release",
+    "package": "cp -r package.json package-lock.json README.md dist",
```
uittorio, line 24: I would like to understand a bit more the need of a package script. We are currently copying these files only for the release. If you were renaming package script to copyPackageAndReadmeToDist I would be okay.

martinjlowm: What about a scope similar to tests: `dist:package` and/or `dist:release` 🤔 ?

uittorio: I'm okay with dist:release :)

uittorio: @martinjlowm Sorry there was a misunderstanding. I think release can stay there. I would do write instead

```
"dist:release": "cp -r package.json package-lock.json README.md dist",
"release": "npm run dist:release && semantic-release",
```

martinjlowm: Morning :) Doneso!

uittorio: completely forgot about this. I'm okay with collect.. I'll merge it :)

Issue #458, opened 2020-08-08
[Need help] Overriding values; Having hard times with a type error (possible issue)


C0ZEN:
### Subject of the issue
I would like to override a mock value but TypeScript is crying about bad typing.  
Nevertheless it seems that what I did is good and should be possible.
So I do not know if this is a logical error or a problem with the `createMock` function.

**Interface:**

```typescript
export interface IFirebaseGuildV2 {
  id?: string| undefined;
  version?: FirebaseGuildVersionEnum.V2 | undefined;
}
```

**Enum:**

```typescript
export enum FirebaseGuildVersionEnum {
  V1 = 1,
  V2 = 2,
}
```

**Test:**

```typescript
let firebaseGuild: IFirebaseGuildV2;

beforeEach((): void => {
  firebaseGuild = createMock<IFirebaseGuildV2>({
    version: FirebaseGuildVersionEnum.V2, // Error here
  });
});
```

Basically I need to be sure that the Firebase guild has a version of `2`.
The model allow `2` or `undefined`.
So telling the mock to override with `2` should be possible from my point of view.

### Your environment
* ts-auto-mock version: 2.3.3
* typescript version: 3.9.7
* node version: 13.7.0
* npm version: 6.14.6

### Expected behavior
No error.

### Actual behavior
```
TS2345: Argument of type '{ version: FirebaseGuildVersionEnum.V2; }' is not assignable to parameter of type 'DeepPartial<IFirebaseGuildV2>'.
Property 'version' is incompatible with index signature.
Type 'FirebaseGuildVersionEnum.V2' is not assignable to type 'DeepPartial<unknown>[] | readonly DeepPartial<unknown>[] | DeepPartial<any> | undefined'.
```

uittorio:
Hi :)! It looks odd. I'll have a look soon. It could be that we are using A wrong implementation of DeepPartial

C0ZEN:
Seems so simple yet it is not working that I just wish for you that this not a brain killer problem 😨 
Should be quite simple to reproduce though.
Reach me if you need testing or more information.
Thanks!

uittorio:
I couldnt reproduce the bug yet, Could you share your tsconfig.json ? Thanks

C0ZEN:
@uittorio here it is.
If you want me you let me know and I can provide a reproduction.

**tsconfig.json:**
```
{
  "compileOnSave": false,
  "compilerOptions": {
    "allowJs": true,
[...]

PR #462, opened 2020-08-09
fix(partial): improve deep partial type

uittorio:
Improve deep partial type to ensure compilation without errors when passing default values to create mock and create mock list.

We could have add 'type-fest' as dependencies in ts-auto-mock. 
Advantages
- we may not have to fix deep partial in the future

Disadvantages
- Increase the bundle size / Add a peer dependency.

I prefer not to add it for now, let's see how it goes.

Partial deep source
https://github.com/sindresorhus/type-fest/blob/master/source/partial-deep.d.ts

- [X] You have added tests


uittorio:
This PR will fix #458

Issue #464, opened 2020-08-09
[Issue] Version 2.3.4 created a type error with the `createMock` function


C0ZEN:
### Subject of the issue
The latest release [2.3.4](https://github.com/Typescript-TDD/ts-auto-mock/releases/tag/2.3.4) related to the issue #458 fixed an error and also created a new one.

### Your environment
* ts-auto-mock version: 2.3.4
* typescript version: 3.9.7
* node version: 13.7.0
* npm version: 6.14.6

### Steps to reproduce

**Test:**

```typescript
client = createMock<Client>({
   user: {
      setPresence: setPresenceMock,
   },
});
```

The new error is on `user`

**Error:**
```typescript
TS2322: Type '{ setPresence: jest.Mock<Promise<Presence>, any>; }' is not assignable to type 'PartialObjectDeep<ClientUser>'.
Types of property 'valueOf' are incompatible.
Type '() => Object' is not assignable to type '(() => string) & (() => Object)'.
Type '() => Object' is not assignable to type '() => string'.
Type 'Object' is not assignable to type 'string'.
```

**Fake Client class:**

```typescript
export class Client {
    public user: ClientUser | null;
}
```

**Fake ClientUser class:**

```typescript
export class ClientUser {
    public setPresence(data: PresenceData): Promise<Presence>;
}
```

### Expected behavior
No error.

### Actual behavior
Type error.

uittorio:
hi @C0ZEN, could you share the type of setPresenceMock?

C0ZEN:
@uittorio if you need all these data I think that the best solution is to give you a complete reproduction.
Note that all types are coming from firebase-admin package though.

uittorio:
I'll just install firebase-admin then!! thanks :P

uittorio:
I've tried to install firebase-admin, but I couldn't find those types.  Sorry to bother you again, I just need the imports references of those. Thanks

C0ZEN:
@uittorio oh sorry I meant discord.js

C0ZEN:
@uittorio the reproduction is available: https://github.com/C0ZEN/ts-auto-mock-issue-464

uittorio:
Cool, thanks @C0ZEN.

The issue is related to the default implementation in an object of valueOf. 

The deepPartial that has been added in 2.2.4 is much smarter and it does consider functions. `objects` have by default valueOf.

[...]

Issue #471, opened 2020-08-15
[Issue] Type instantiation is excessively deep and possibly infinite while using `createMock`


C0ZEN:
### Subject of the issue

This is related to a type issue when using the `createMock` function with discord.js.
They landed a new version [12.3.0](https://github.com/discordjs/discord.js/releases/12.3.0) and there is an issue visible on [this project](https://github.com/Sonia-corporation/il-est-midi-discord/pull/758).

**Test:**

`createMock<Client>()`

Client is from `discord.js`.

**Error:**

`Type instantiation is excessively deep and possibly infinite.`

### Your environment
* ts-auto-mock version: 2.3.4
* typescript version: 3.9.7
* node version: 13.7.0
* npm version: 6.14.6

### Expected behavior
No error.

### Actual behavior
Type error.

uittorio:
Hi @C0ZEN. Sorry to get back to you late, I had a busy week. I am looking into this issue and https://github.com/Typescript-TDD/ts-auto-mock/issues/464

C0ZEN:
@uittorio no problem this is not blocking on my side so take your time.

uittorio:
It looks like there is an issue on Typescript. I am not sure it is the same issue but It has been added to typescript 4.0.1 milestone https://github.com/microsoft/TypeScript/issues/34933.

uittorio:
I've noticed that without the casting everything is working

C0ZEN:
@uittorio nice catch.
As I said I am not in a hurry to update discord.js but can you tell me what is the best option here for me?
Wait that you come up with a solution - with the TypeScript error I suppose you will wait -, wait for TypeScript to come up with a solution or just add the `valueOf` like above?

uittorio:
https://github.com/Typescript-TDD/ts-auto-mock/issues/464#issuecomment-674516498 I think this would be the best approach if you want to update to the new discord version

uittorio:
This should be fixed now by 2.3.5.. @C0ZEN feel free to close it once you get the chance to confirm that is working :)

C0ZEN:
@uittorio I will give it a try tonight and if all good I will close this PR.
Thanks for the fix.

PR #474, opened 2020-08-16
fix(createMock): PartialDeep to ignore object prototype properties

Pmyl:
Change order of conditional typings in PartialDeep to enable overriding Object prototype properties.

Don't ask me how or why it works, it just does. I would be curious to try this fix in a repo that uses ts-auto-mock estensively.

Also fixed es-lint to allow windows-style new-lines. When pushing it will transform everything to unix style anyway.

This should fix the issue #464 and consequentially #471


src/partial/partial.ts
```diff
@@ -9,11 +9,11 @@ export type PartialDeep<T> = T extends Primitive
   : T extends ReadonlyMap<infer KeyType, infer ValueType>
   ? PartialReadonlyMapDeep<KeyType, ValueType>
   : T extends ReadonlySet<infer ItemType>
-  ? PartialReadonlySetDeep<ItemType> // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  : T extends (...args: any[]) => unknown
-  ? T | undefined
+  ? PartialReadonlySetDeep<ItemType>
   : T extends object
   ? PartialObjectDeep<T>
+  : T extends (...args: any[]) => unknown // eslint-disable-line @typescript-eslint/no-explicit-any
```
uittorio, line 15: Considering that a function is always an object, shall we remove this completely?

I mean 

```
: T extends (...args: any[]) => unknown // eslint-disable-line @typescript-eslint/no-explicit-any
? T | undefined
```

uittorio: The only problem with this approach, removing or moving the line up is this scenario.

```ts
type B = () => string;

  const b: PartialDeep<B> = {
    a: 's',
  };
```

This doesn't error. If you instead put back the function check it will error
```
Type '{ a: string; }' is not assignable to type 'B'.   Object literal may only specify known properties, and 'a' does not exist in type 'B'.
```

Pmyl: You're right, I'll try to find another way

Pmyl:
New commit new try.
This problem is not easy to resolve, we can't tell typescript/javascript that when we do `{ a: 1 }` we don't want valueOf. We do get it, hence typescript says "valueOf is of a different type than the one specified in the discord library".

What we can do is to hack it a bit to relax the type and make it easy to use createMock values even in these cases.
**It's still a hack though, and it has pro and cons.**
The only way I've found to overcome this problem is to... ignore the overrides and always use the types from Object for the prototype properties. I know, it's not great, but at least it works 99.9% of the time.

This causes two effects:
**If the override changes the type completely it will fail when applying values in createMock**
```ts
interface Int {
  a: string;
  toString(): number;
}
createMock<Int>({
  toString: () => 123
  // ERROR, Type 'number' is not assignable to type 'string'.
  // because since it ignores the override it means that toString
  // has the type defined in Object: () => string
})
```

**The type of valueOf in Object is () => Object, meaning that even if the interface overrides it to () => string you can still return whatever you want (because Object matches everything apparently)**
```ts
interface Int {
  valueOf(): number;
}
createMock<Int>({
  valueOf: () => "123"
  // This is valid because PartialDeep ignores the override
  // and this matches the type specified in Object for valueOf: () => Object
})
```

Last thing, let's remember to squash, I've gone through various stages of investigation and I have too many commits.

Let me know what you think @uittorio

uittorio:
Okay, I finally made up my mind about this functionality. I personally don't like the workaround that we are putting in place in the deepPartial but this solution will not break the common usage of createMock, it will only break if you are trying to mock prototype properties of an Object (really odd if you are doing that). Happy with the changes.

uittorio (approved):

PR #492, opened 2020-08-27
WIP: feat(debug): enhance the logs to add the name of the unsupported type 

C0ZEN:

- [ ] Your code builds clean without any errors or warnings
- [ ] You have added tests


C0ZEN:
@uittorio hi.
I do not know how to find the name of the property for the `typeNotSupported` method.
I never saw this kind of TypeScript before.
I suppose this is something you already coded so could you point me in the right way please?

Pmyl:
You can try to use `PropertySignatureCache.instance.get()`, this method will return the PropertyName of the property that the transformer is currently processing. In case the transformer is currently processing the initial type (MyInterface in case of `createMock<MyInterface>()`) it will return an empty PropertyName that should result in empty string.

Once you have the PropertyName you can use `TypescriptHelper.GetStringPropertyName(propertyName)` to get the name as a string.

You can see it in action in the file `src/transformer/descriptor/string.ts`.

C0ZEN:
@Pmyl I changed my code to reflect the example you gave me but there is no test created yet for this case and I am unable to guarantee that this information (property name) is the right one and that it will be available for the three callers of this method.

Though the two worst cases are:
- it will change nothing to the logs if the property name is not available (I made sure of that with the checks)
- it will display not the thing I expected

Please review this and tell what do you think.

uittorio:
@C0ZEN I was thinking about printing the current node. Let's say we don't support `StringLiteral`

interface Hello { property: string }

createMock<Hello>() 

Te out output will be

```Not supported type: StringLiteral: string - it will convert to null``` 

I have to try this out but there is a  file `printNode.ts` that contains a function that should give us a human-readable node

uittorio:
@C0ZEN thanks for bringing this to our attention. I would like to improve this functionality further. I may try to add  

- the file test where the create mock fail
- the type that is failing (Activity for example)
- the node kind (we are already logging this)
- the property, if present, that has the type that fails.

I should be able to do something in the next couple of days :)

C0ZEN:
@uittorio yeah this is wonderful if you are able to do it all. 😃 
If you could include the file test as a link (like when Jest show an error) this could be huge.

Also I suggest you to finish this PR with only a name provided and open another PR for the new changes. It will close all related issues and avoid a huge PR. 🤔

uittorio:
@C0ZEN I agree, I want to finish a POC to make sure we are going in the right direction for the logging. We will definitely create small PR for each feature. I am also planning to improve our testing to be able to check the error we log when something goes wrong. 

No problem with the docs error!

I'll let you know. In the meantime, if you want you can open a PR with everything except the logging stuff. Up to you! :)

C0ZEN:
@uittorio I added small PR to close other issues.

uittorio:
I'm working on a PR to enhance our tests suite to be able to check if a warning has been logged.

uittorio:
Hi @C0ZEN I took the liberty to create another PR to enhance our logs. Your input was fundamental to understand what to do. [This is the PR. You can see the logs](https://github.com/Typescript-TDD/ts-auto-mock/pull/513)

Basically we just add file link to the createMock and link to the error. If you know how to remove file:// from the link would be great :P

C0ZEN:
Done in 513.

PR #513, opened 2020-09-06
feat(logs): enhance unsupported type logs

uittorio:
Enhance unsupported type logs by adding source file where the mock is created and source file where the type is failing

What did I change?
 - Every-time the compiler runs the tsAutoMock.log file will be cleared. winstonFileLogger is not a singleton and it needs this options to make sure it will reset the file the first time it is created.
```
options: { flags: 'w' },
```
- I've added a couple of tests to make sure we are logging the right informations.
- the new test logs sections has the cache disabled. I've decided to disabled it to make it easier to check the warnings. With the cache enabled I would have to duplicate the conditional type between createMock and createMockList. This is quite important. We don't log another warning if it was part of an existing mock. :) 

Logs examples.

```
2020-09-06T18:29:41.029Z - WARNING: Transformer - Not supported type: ConditionalType - it will convert to null
created file:///Users/vittorioguerriero/Documents/GitHub/ts-auto-mock/test/logs/createMock.warning.test.ts:15:5
used by file:///Users/vittorioguerriero/Documents/GitHub/ts-auto-mock/test/logs/conditionalType.ts:1:34
2020-09-06T18:29:41.194Z - WARNING: Transformer - Not supported type: ConditionalType - it will convert to null
created file:///Users/vittorioguerriero/Documents/GitHub/ts-auto-mock/test/logs/createMockList.warning.test.ts:15:5
used by file:///Users/vittorioguerriero/Documents/GitHub/ts-auto-mock/test/logs/conditionalType.ts:1:34
```

- [x] You have added tests


src/transformer/descriptor/indexedAccess/indexedAccess.ts
```diff
@@ -33,7 +33,8 @@ export function GetIndexedAccessTypeDescriptor(
         default:
           TransformerLogger().typeNotSupported(
             'IndexedAccess of TypeReference of ' +
```
C0ZEN, line 35: Could use template literal instead of concatenation

src/transformer/descriptor/indexedAccess/indexedAccess.ts
```diff
@@ -44,7 +45,8 @@ export function GetIndexedAccessTypeDescriptor(
       break;
     default:
       TransformerLogger().typeNotSupported(
-        'IndexedAccess of ' + ts.SyntaxKind[node.indexType.kind]
+        'IndexedAccess of ' + ts.SyntaxKind[node.indexType.kind],
```
C0ZEN, line 48: Could use template literal instead of concatenation

C0ZEN (dismissed):

test/logs/utils/log.ts
```diff
@@ -0,0 +1,29 @@
...
+  const logErrorFile: string = fs.readFileSync('./tsAutoMock.log', 'utf-8');
+  const logs: UnsupportedTypeLog[] = [];
+  const lines: string[] = logErrorFile
+    .split('\n')
+    .filter((line: string) => !!line);
+
+  for (let i: number = 0; i < lines.length; i += 3) {
+    logs.push({
+      header: lines[i],
+      created: lines[i + 1],
```
Pmyl, line 21: This is extremely opinionated in a way that if we have any other way to write warnings (for example a one line warning in style "you can't use that while using random feature") we have to update this function... I think I'm ok with it, I just wanted to point it out

Pmyl (approved):
Looks good to me, if you want to address the comments about string concatenation or not I'm ok whatever

src/transformer/descriptor/indexedAccess/indexedAccess.ts
uittorio: Ill do that :) thanks!

test/logs/utils/log.ts
uittorio: I agree. I would say that once we need it we can improve the utility to get the logs :)

uittorio: Another way to do it would be to parse the warnings based on the service name. For example ... divide the logs starting from WARNING - {serviceName}, nothing to difficult.

PR #519, opened 2020-09-10
feat(definitelyTypes): do not log warnings generated in files out of processed directory

Pmyl:
Reduce the number of types that generate warnings by excluding false positives.

### DIFF
--- a/definitelyTypedTests/index.js
+++ b/definitelyTypedTests/index.js
@@ -133,7 +133,7 @@ createDefinitelyTypedMock<typeof pak>();
                 console.warn('☐');
                 outputService.addData(processId, dir, {
                     response: 'warning',
-                    message: response.toString()
+                    message: removeWarningLogsForFilesInDifferentDirectories(response.toString(), typePath)
                 });
             } else {
                 process.stdout.write(`TYPE: ${dir} P${processId} `);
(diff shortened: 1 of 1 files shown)


definitelyTypedTests/index.js
```diff
@@ -161,3 +161,19 @@ createDefinitelyTypedMock<typeof pak>();
...
 }
+
+function removeWarningLogsForFilesInDifferentDirectories(log, typePath) {
+    const logLines = log
+        .split('\n')
+        .filter((line) => !!line);
+
+    for (let i = logLines.length - 1; i >= 0; i--) {
+        if (/WARNING[^\n]+?: Transformer/.test(logLines[i])) {
+            if (!logLines[i+2].includes(typePath)) {
```
uittorio, line 172: If I understood the problem correctly we would like to remove definitely typed errors from other definitely type dependencies? I think there is still the possibility that a definitely typed depends on NON definitely typed types. What would happen in that case?

Pmyl: Two alternatives:
- if the package is installed in the definitely typed node_modules then it will still work since the file is in the folder `DefinitelyTyped/types/CURRENT_TYPE/node_modules/***`
- if the package is installed in the root folder then it won't work

You're right... you know what, I'll close this PR, I want to do something different with the warnings in the UI.

PR #526, opened 2020-09-13
fix(enum): fix enum constant computed properties

Pmyl:
- Fixed case of constant enum computed member:
```ts
enum Constant {
  Value = 1 + 2
}
```
- added back npm script to run tests on the built version of playground, this will enable debugging the emitted code

Unfortunately non-constant enum computed members are still not fixed, to do that we have to find once and for all a way to import the enum in the file of the mock and use that directly.
Example of non-constant enum computed member:
```ts
function nonConstant(): number {
  return Math.Random();
}
enum Computed {
  Value = nonConstant()
}
```

### DIFF
--- a/test/features/random/enum.test.ts
+++ b/test/features/random/enum.test.ts
@@ -96,15 +96,9 @@ describe('Random enum', () => {
     expect(thirdMock.direction).toBe(Direction.None);
   });
 
-  it('should have position of "computed" property, because transformer don\'t support computed Enum values', () => {
-    // Issue https://github.com/Typescript-TDD/ts-auto-mock/issues/339
-    function getComputed(): number {
-      return 12;
-    }
-
-    enum WithComputed {
-      Computed = getComputed(),
-      Computed2 = 1 + 4,
+  it('should get the correct const enum member computed property', () => {
+    const enum WithComputed {
+      Computed = 1 + 4,
     }
 
     interface IWithComputed {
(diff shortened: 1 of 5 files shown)


test/features/random/enum.test.ts
```diff
@@ -96,15 +96,50 @@ describe('Random enum', () => {
...
+
+    const spy: jasmine.Spy = spyOn(Math, 'floor');
+
+    spy.and.returnValue(0);
+    const mock1: IWithComputed = createMock<IWithComputed>();
+
+    expect(mock1.computed).toBe(5);
+  });
+
+  it('should get the correct enum member computed property for constant expression', () => {
```
uittorio, line 116: Is this one a duplicated test? Looks like the one above.

Pmyl: `const enum` vs `enum`

uittorio: Ignore my comment

test/features/random/enum.test.ts
```diff
@@ -96,15 +96,50 @@ describe('Random enum', () => {
     expect(thirdMock.direction).toBe(Direction.None);
   });
 
-  it('should have position of "computed" property, because transformer don\'t support computed Enum values', () => {
+  it('should get the correct const enum member computed property', () => {
```
uittorio, line 99: Do you think we should add a test outside the random functionality?

Pmyl: done

uittorio (approved):

PR #527, opened 2020-09-13
fix(definitelyTyped): do not fail on mocking module declarations

Pmyl:
Do not fail on mocking module declarations, this reduces the amount of errors and warning in definitely typed tests


config/modules/definitelyTypedTransformer/webpack.functions.js
```diff
@@ -1,10 +1,12 @@
 const path = require("path");
 const { merge } = require("webpack-merge");
 const base = require("../base/webpack.base");
+const DetermineDevToolFromEnvironmentDebugMode = require("../../utils/devtool");
 
 module.exports = merge(base({
     tsConfigFile: 'config/modules/definitelyTypedTransformer/tsconfig.json'
 }), {
+    devtool: DetermineDevToolFromEnvironmentDebugMode(),
```
uittorio, line 9: We could move this one in the base webpack.

Pmyl: I agree, I'll do it

Pmyl: done

ui/src/definitely-typed/filters/definitelyTypedFilters.tsx
```diff
@@ -2,80 +2,174 @@ import React, { useEffect, useState } from 'react';
...
+      );
+    }
+  );
+
+  type PresetFilterOption = PresetFilter & { disabled?: never; };
+
+  let errorPresetFilters: PresetFilterOption[] = [];
+
+  if (isShowing === 'error') {
+    errorPresetFilters = errorPresetFilters.filter(filter => props.data.some(d => filter.regex.test(d.message || '')));
```
uittorio, line 110: I think you wanted to use ErrorPresetFilters instead of errorPresetFilters

```
errorPresetFilters = ErrorPresetFilters.filter(filter => 
```

ui/src/definitely-typed/filters/definitelyTypedFilters.tsx
```diff
@@ -2,80 +2,174 @@ import React, { useEffect, useState } from 'react';
...
-            isShowingWarnings: isShowing === 'warning',
-            isShowingSuccesses: isShowing === 'success'
-        };
-    }
+export function DefinitelyTypedFilters(
+  props: DefinitelyTypedFiltersProps
+): JSX.Element {
+  const [filterOut, setFilterOut] = useState('');
+  const [filterIn, setFilterIn] = useState('');
+  const [errorPresetFilterId, setErrorPresetFilterId] = useState(0);
```
uittorio, line 35: I think it's going to be difficult in the future to understand. Shall we convert to an enum at least? None = 0, Custom = 1 and we don't care about the rest

ui/src/definitely-typed/filters/definitelyTypedFilters.tsx
```diff
@@ -2,80 +2,174 @@ import React, { useEffect, useState } from 'react';
...
+  }
+
+  function buildFilterInOption() {
+    if (isShowing === 'error') {
+      return errorPresetFilterId === 1 && filterIn
+        ? new RegExp(filterIn, 'i')
+        : errorPresetFilterId === 0
+        ? null
+        : errorPresetFilterId !== 1
+        ? errorPresetFilters[errorPresetFilterId - 3].regex
```
uittorio, line 63: If you want to avoid the  -3 we could create the filter options first with all the values, then create the React Element from that list.

uittorio (changes requested):
Minor changes.

ui/src/definitely-typed/filters/definitelyTypedFilters.tsx
Pmyl: done, I'm not sure what happened here...

ui/src/definitely-typed/filters/definitelyTypedFilters.tsx
Pmyl: created an enum with -1 and -2 as values so we don't have to do -3 in the calculations

ui/src/definitely-typed/filters/definitelyTypedFilters.tsx
Pmyl: addressed, with the enum the -3 is not anymore needed

uittorio (approved):

PR #530, opened 2020-09-14
fix(properties): create property when undefined/void

Pmyl:


src/transformer/descriptor/mock/mockProperties.ts
```diff
@@ -44,6 +44,10 @@ export function GetMockPropertiesFromDeclarations(
         return false;
       }
 
+      if (member.questionToken) {
```
uittorio, line 47: This is cool. I think we can safely remove the code below from file propertySignature.ts

```
if (node.type) {
    if (node.questionToken) { // remove this if
      return GetUndefinedDescriptor();
    }

    return GetDescriptor(node.type, scope);
  }
```

What do you think?

test/playground/playground.test.ts
```diff
@@ -10,9 +10,14 @@ import { createMock } from 'ts-auto-mock';
```
uittorio, line 10: revert this one if you don't mind.

uittorio (changes requested):
One comment!! Nice one!

src/transformer/descriptor/mock/mockProperties.ts
Pmyl: it makes sense, we exclude it early, it should never get to that point if it has the question token, I'll remove it later today

Pmyl: done

test/playground/playground.test.ts
Pmyl: done

uittorio (approved):

PR #542, opened 2020-09-16
fix(transformer): do not fail when accessing index of extended this for a computed property

Pmyl:
This will reduce errors also in definitely typed tests.


test/logs/indexedAccess/indexedAccess.warning.type.ts
```diff
@@ -0,0 +1,7 @@
+export interface OtherInterface extends StandardInterface {
+  '[]': string;
+}
+
+export interface StandardInterface {
```
uittorio, line 5: Wow....I don't really understand this scenario...

I wrote the same interface but I've added a property key in the input of the method`prop`
If I use OtherInterface, I can write something like this

```ts
export interface OtherInterface extends StandardInterface {
  '[]': string;
}

export interface StandardInterface {
  b: number;
  prop<K extends keyof this>(k: K): this[K];
}

export class Hello implements OtherInterface {
  public b: number;
  public '[]': string;

  constructor() {
    this.b = 123;
  }

  public prop<K extends keyof this>(k: K): this[K] {
    return this[k];
  }
}

const b: string = new Hello().prop('[]');
const c: number = new Hello().prop('b');
```

Technically we should be able to infer the type.

uittorio: Can you add in the list of unsupported type this specific scenario or it is already covered?

uittorio (approved):

test/logs/indexedAccess/indexedAccess.warning.type.ts
Pmyl: It's already covered [here](https://typescript-tdd.github.io/ts-auto-mock/types-not-supported#indexed-access-type-with-generics)... it's slightly more convoluted but it's the same thing: indexing a property or a type using a generic that extends keyof type. The only difference is that in this case the type is `this` and the property name is computed.

This PR fixes the case where the transformer was failing when the property name is computed, now it gives a warning in the same way it does for non computed property names.

PR #569, opened 2020-09-18
various fixes

Pmyl:
Three commits:

`chore(definitelyTyped): fix case when module has default unnamed export`
`chore(docs): add 'other' option when filtering definitely typed errors`
`fix(transformer): apply null and warning when property type cannot be identified`

I won't squash them in one


src/transformer/descriptor/module/module.ts
```diff
@@ -134,7 +134,8 @@ export function GetPropertiesFromSourceFileOrModuleDeclaration(
         }
 
         return TypescriptCreator.createPropertySignature(
-          d.originalDeclaration.name as ts.Identifier,
+          (d.originalDeclaration.name as ts.Identifier) ||
+            ts.createIdentifier('default'),
```
uittorio, line 138: Do you still need the casting?

uittorio (approved):

src/transformer/descriptor/module/module.ts
Pmyl: yes, apparently `name` can be also a "DeclarationName" that doesn't match "Identifier"

PR #612, opened 2020-12-20
Feature hydrated mocks

uittorio:
Hi, this PR tentative adds createHydratedMock as a new mock function that will allow developers to ignored optional properties.

Example
```ts
interface Interface {
    notRequired?: string;
}

const mock: Interface = createHydratedMock<Interface>();
expect(mock.notRequired).toBe('');
```

**This PR is not final**. I've opened it to show that we need a better design for this functionality. 
In order to add createHydratedMock, I've added a new DeclarationCache, FactoryCache, and a FactoryRegistrationPerFile. This approach is quite complicated because every time we refer to the DeclarationCache we need to decide which cache to use.

Here is the list of the functionality that relies on DeclarationCache
- createMock 
- registerMock
- enum
- generics
- typeParameters (generics)

@Pmyl What do you think? I am sure you had a similar, hopefully better idea.

If you think it's fine to have different caches I'll try to make the code easier to extend in the future :) 

Update 25/12/2020
I've updated the code to support all the scenarios.
I would like to refactor the mockDefiner because there is some duplication. I'll do it if this PR gets approved


Pmyl (approved):
I think the mock definer can be refactored a bit more now that you've done your refactor, I think the next step is to do that for the next PR... when I'm not lazy

this one looks good though 👍

PR #631, opened 2021-03-20
Fix setting env variable

uittorio:

- [ ] You have added tests


uittorio:
Github actions breaking changes on set-env. There was a security issue
. https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

This is the new approach:
https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

PR #736, opened 2021-05-16
fix(hydratedMocks): fix recursion

uittorio:
make sure interface declaration and class declaration are returning the original node when using the getTypes function so the union functionality can use the returned value without encountering maximum call stack


src/transformer/descriptor/type/type.ts
```diff
@@ -41,7 +41,20 @@ export function GetType(node: ts.Node, scope: Scope): ts.Node {
...
     );
 
-    return GetType(declaration, scope);
+    const type: ts.Node = GetType(declaration, scope);
+
+    // if the type is a reusable and finite (reusable types except for typeAlias)
+    // we can return the original node
+    // so it can be re used and we handle recursion
+
+    if (
```
Pmyl, line 50: does it mean that in case of an object type it will still give error?

```
  type B = A;

  type A = {
    recursiveProp: B | null;
    aProp: string | null;
  };
  
  createHydratedMock<A>();
```

uittorio: Good point! Maybe I should check if its an object like :)

uittorio: Done :)

Pmyl (approved):

PR #740, opened 2021-05-18
feat(getAccessor): correctly mock get accessors for mocked classes

Pmyl:

- [x] You have added tests


Pmyl:
I have a question for you @uittorio , do I have to use the property cache? Since a getter is a property I think I should use it but I don't remember what that for

uittorio:
That is correct! PropertyCache responsibility is to store the name of the property and consume it immediately after, it was a pain to pass it around

Pmyl:
> That is correct! PropertyCache responsibility is to store the name of the property and consume it immediately after, it was a pain to pass it around

Just to understand the functionality better, why don't we use it to store method names as well?

Pmyl:
Also, what test can I write to make sure a future me doesn't remove the cache usage?

uittorio:
I need to double check but Methods and properties on classes and interfaces are using the propertyCache functionality. You can see it from the GetProperties function.

How do you test it?
the framewework test suite checks that the spies with jasmine have the correct names. You could extend those tests with getter and setter and is if that works :)

uittorio:
Apparently it is also used when we generated the random string values.

src/transformer/descriptor/getAccessor/getAccessor.ts
```diff
@@ -0,0 +1,20 @@
...
+import { GetFunctionReturnType } from '../method/functionReturnType';
+
+export function GetGetAccessorDeclarationDescriptor(
+  node: ts.GetAccessorDeclaration,
+  scope: Scope
+): ts.Expression {
+  const returnTypeNode: ts.Node = GetFunctionReturnType(node);
+  const returnType: ts.Expression = GetDescriptor(returnTypeNode, scope);
+
+  if (!node.name) {
```
uittorio, line 13: Why do we need this check?

uittorio (commented):
Nice one!! Are you also doing the setter? :100:

Pmyl:
@uittorio doing the setter may be a problem, what if we have both getter and setter? Is that going to break the functionality due to the fact that we're mocking the same property twice? I'll write a test to see if everything works fine

src/transformer/descriptor/getAccessor/getAccessor.ts
Pmyl: removed this check

Pmyl:
@uittorio I've done the changes to handle setters as well, in essence I removed the "not supported" warning by returning undefined and handled case when setter was defined before getter

uittorio (approved):

PR #770, opened 2021-06-03
refactor(transformer): make creation of mock configurable and shift responsibility to custom functions to identify configuration

Pmyl:
Moved responsibility to identify options to the custom functions, option interface used to change the created mock, this way it's easier to create new custom functions with different mix of options (for example the missing hydrated-list).

This is also a prerequisite for the builder feature and the depth limit feature

### DIFF
--- a/src/transformer/mock/mock.ts
+++ b/src/transformer/mock/mock.ts
@@ -33,79 +33,84 @@ import {
 } from '../../typescriptFactory/typescriptFactory';
 import { SetCurrentCreateMock } from './currentCreateMockNode';
 
-function getMockExpression(nodeToMock: ts.TypeNode): ts.Expression {
-  return GetDescriptor(nodeToMock, new Scope());
-}
-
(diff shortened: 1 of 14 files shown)


src/transformer/mock/mock.ts
```diff
@@ -33,79 +33,84 @@ import {
...
+  const mockExpression: ts.Expression = getMockExpression(options);
 
-  const mockExpression: ts.Expression = getMockExpression(nodeToMock);
-
-  if (hasDefaultValues(node)) {
-    return getMockMergeExpression(mockExpression, node.arguments[0]);
+  if (!!options.amount) {
+    return getListOfMocks(
+      mockExpression,
+      options as CreateMockOptions & { amount: ts.Expression }
```
uittorio, line 53: strange that you have to cast it after the type guard check !!options.amount.

uittorio (approved):
I like it!! Basically you've improved the existing design from a simple configuration to a behavioural configuration :) Really nice!

PR #802, opened 2021-06-22
feat(*): export alternative transformer function to allow providing customized typescript

Pmyl:
Small step to get to ts-auto-mock playground in the docs, now we can provide a custom typescript version at runtime before running the transformer.

To achieve it I've created a new transformer exported in the `transformer/index.ts`.
Added a test to cover that nothing (other than transformer.ts) uses typescript from the import ensuring that the whole transformer only uses the provided version.


src/transformer/typeReferenceReusable/typeReferenceReusable.ts
```diff
@@ -1,11 +1,18 @@
...
+function getReusableTypes(): ts.SyntaxKind[] {
+  return [
+    core.ts.SyntaxKind.ClassDeclaration,
+    core.ts.SyntaxKind.InterfaceDeclaration,
+    core.ts.SyntaxKind.TypeAliasDeclaration,
+  ];
+}
 
 export function isTypeReferenceReusable(declaration: ts.Declaration): boolean {
+  reusableTypes = reusableTypes || getReusableTypes();
```
uittorio, line 15: You could just inline the reusableTypes inside the function instead of a method. It's true that your way caches the array... but I don't think there is a performance concern here

Pmyl: makes sense, done

uittorio (approved):

PR #905, opened 2021-09-06
Feat arrow functions types

uittorio:
Hey, this is the second part of arrow functions without types. I've added some scenarios. 

In ts-auto-mock, most of the stuff works with declarations. Usually, you start from a node and:
-  you get a type
- you get a symbol
- you get a declaration
- and you send it over the descriptor again!

For non typed functionalities, there is no declaration. The alternative that I've found is to use `typeChecker.typeToTypeNode`, it's a great functionality but you cannot use the typeCheker on these nodes after doing that.

I've created a small function that checks and alerts when the final type cannot be checked again. Basically, this will work for primitive types in our code base. For example
```ts
 SyntaxKind.NumberKeyword,
  SyntaxKind.StringKeyword,
  SyntaxKind.BooleanKeyword,
  SyntaxKind.ArrayType,
  SyntaxKind.UndefinedKeyword,
```

We are only using this functionality when  FunctionLikeDeclaration doesn't have a return body defined, it may be handy for something else.


uittorio:
@Pmyl your typescript import checked worked!! the test are failing for that reason!! wowowowowow! I'll fix it later but if you want to check the rest it would be great

uittorio:
@Pmyl ready :)

src/transformer/descriptor/descriptor.ts
```diff
@@ -46,6 +46,12 @@ import { GetShorthandPropertyAssignmentDescriptor } from './shorthandPropertyAss
 import { GetParameterDescriptor } from './parameter/parameter';
 import { GetVariableDeclarationDescriptor } from './variable/variable';
 
+const GetParenthesizedExpressionDescriptor: (
```
Pmyl, line 49: is there a reason why this is not in its own file? I don't mind but I want to make sure it's not a mistake

src/transformer/descriptor/typeNode/convertNodeToTypeNode.ts
```diff
@@ -0,0 +1,49 @@
...
+    undefined,
+    undefined
+  );
+
+  if (!typeNode) {
+    TransformerLogger().typeCannotBeChecked(node);
+    return GetNullDescriptor();
+  }
+
+  if (!canTypeBeChecked(typeNode)) {
```
Pmyl, line 24: if the type can or cannot be checked the result is the same, is this useful?

uittorio: Probably I need to return a NullDescriptor.  :) thanks! I'll see if I can write a test.

src/transformer/descriptor/descriptor.ts
uittorio: Just a mistake :) forgot to refactor!

src/transformer/descriptor/typeNode/convertNodeToTypeNode.ts
uittorio: Done, cannot find a test

src/transformer/descriptor/descriptor.ts
uittorio: done

Pmyl (approved):
nice

PR #1126, opened 2022-05-02
build(dependencies): node globals types are now excluded when analyzing type properties

uittorio:
Node 16.6 introduced Array.at -> https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V16.md#2021-07-29-version-1660-current-bethgriggs
As a consequence of that @types/node 16.6 introduced the types for that
https://github.com/DefinitelyTyped/DefinitelyTyped/commit/dca7357e90ef804954aaf042ed158220bcd09079#diff-2696281f8832b1227bb41483a7525a8ccd9399bb4b1704f69c8c54bb1346d8edR80


uittorio:
I forgot to mention the reason I've made this change. A few tests started failing because typescript merges all the type declarations and some of the tests were expecting the new method 'at'

There is one scenario that I could not reproduce. If you use for example type myType = Array<string>, would that go to typescript/lib and ignore it or would it go to node globals first?
Only the tests about extending an Array were failing.

src/transformer/descriptor/nodeTypes/nodeTypes.ts
```diff
@@ -0,0 +1,14 @@
+import type * as ts from 'typescript';
+
+const NodeTypesGlobalFile: string = 'node_modules/@types/node/globals.d.ts';
+
+export function IsNodeType(node: ts.Node): boolean {
```
Pmyl, line 5: Maybe it's better if we specify "nodejs" to be clear we're not meaning ts.Node

Pmyl (commented):
Just one comment

src/transformer/descriptor/nodeTypes/nodeTypes.ts
uittorio: @Pmyl I've made the changes. 

I've also updated the commit message to include the change in a version, this is a fix/feature. I would say fix. I don't think we want to mock any node js globals types ever...

Pmyl (approved):

PR #1411, opened 2022-09-20
docs(ui): archive repository

uittorio:

- [ ] Your code builds clean without any errors or warnings
- [ ] You have added tests


Pmyl (approved):
Sad but necessary. If the typescript team decides to create an easy way to include transformers in projects we can think of restarting development... as long as the typescript world doesn't become even more fragmented than it is right now (babel, tsnode, swc, esbuild, etc...).

Approved

uittorio:
@Pmyl thanks for the review, I was actually reading about archiving repositories on github and it seems like it will be readonly, dependabot will be disabled as well. This means that we are not going to be able to do updates. Are we still happy with that, if so, I will update the description accordingly. Another approach we could take is that we don't archive it yet but we just leave a message.

Pmyl (approved):

